PR #41, opened 2022-09-29
Run build script before packing the npm package

akloboucnik:
This is a follow-up of #40 - what I assume happened is that before publishing version `0.0.4` to npm registry, `npm build` was not run so `0.0.4` actually does not contain declaration files.

This ensures it's always run before the package is created (using `npm pack`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "rabbitmq-stream-js-client",
-  "version": "0.0.4",
+  "version": "0.0.5",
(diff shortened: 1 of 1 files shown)


gpad:
Hi @akloboucnik, thanks for the PR. We are working on automatically publishing the npm package via CI #49, so this PR is not necessary anymore. Thanks again, and feel free to contribute again. We are working on the project to be compliant with other clients so we can publish and consume messages ASAP.

PR #42, opened 2022-10-18
fix username duplicity

jzfre:
there is obviously a bug causing inability to connect to rabbitmq instances where the username password is not the same


ejf-coders51 (approved):
tnx

PR #46, opened 2022-12-28
close command implementation

baldarn:


src/connection.ts
```diff
@@ -75,10 +77,15 @@ export class Connection {
...
-  public close(): Promise<void> {
+  public async close(
+    params: { closingCode: number; closingReason: string } = { closingCode: 0, closingReason: "" }
+  ): Promise<void> {
     this.logger.info(`Closing connection ...`)
     this.heartbeat.stop()
-    return new Promise((res, _rej) => this.socket.end(() => res()))
+    this.logger.debug(`Close ...`)
+    const res = await this.sendAndWait<CloseResponse>(new CloseRequest(params))
+    this.socket.end()
```
gpad, line 87: You have to return when the `end` is completed, see the old code:

```typescript
    return new Promise((res, _rej) => this.socket.end(() => res()))
```

test/e2e/basic_publish.test.ts
```diff
@@ -32,7 +32,6 @@ describe("publish a message", () => {
     await eventually(async () => {
       expect((await rabbit.getQueueInfo(stream)).messages).eql(1)
     }, 10000)
-    await connection.close()
```
gpad, line 35: Why do you remove this code?

test/e2e/basic_publish.test.ts
```diff
@@ -47,6 +46,5 @@ describe("publish a message", () => {
     await eventually(async () => {
       expect((await rabbit.getQueueInfo(stream)).messages).eql(100)
     }, 10000)
-    await connection.close()
```
gpad, line 50: Why do you remove this code?

test/e2e/basic_publish.test.ts
baldarn: the
```
afterEach(() => connection.close())
```

closes the connection 2 times, giving an error now that the close on the rabbit level is done. do you think we have to handle that differently?

test/e2e/basic_publish.test.ts
baldarn: the
```
afterEach(() => connection.close())
```

closes the connection 2 times, giving an error now that the close on the rabbit level is done. do you think we have to handle that differently?

test/e2e/basic_publish.test.ts
gpad: mmh you are right I think it's fine in this way

test/e2e/basic_publish.test.ts
gpad: as above

gpad (approved):

PR #47, opened 2022-12-29
Add ability to encode message in AMQP 1.0

gpad:


Yattol (commented):
LGTM, but an additional look from a different pair of eyes would be wise

albertobarrila (approved):
looks good to me

PR #49, opened 2023-01-04
added automatic publication of npm package

baldarn:


baldarn:
TODO: before merging the secret `${{secrets.NPM_TOKEN}}` must be set ;)

package.json
```diff
@@ -10,7 +10,9 @@
     "check-ts": "tsc --noEmit && tsc --project test/tsconfig.json --noEmit",
     "check-lint": "tsc --noEmit && eslint 'src/**/*.ts' 'test/**/*.ts'",
     "check-format": "prettier -c 'src/**/*.ts' 'test/**/*.ts'",
-    "check-spell": "cspell 'src/**/*ts' 'test/**/*ts'"
+    "check-spell": "cspell 'src/**/*ts' 'test/**/*ts'",
+    "types:prepare": "rm -rf dist && tsc --declaration true --sourceMap false && cp package.json dist/package.json && cp README.md dist/README.md && cd dist && npm version $(git describe --tags --always)",
```
gpad, line 14: Insetad of `tsc --declaration true --sourceMap false` why don't use the build?

package.json
```diff
@@ -10,7 +10,9 @@
...
+    "types:publish": "cd dist && npm publish --access public"
```
gpad, line 15: I have always published the entire package not only dist. Are you sure it's a good idea publish only `dist`

gpad:
What do you think about this PR https://github.com/coders51/rabbitmq-stream-js-client/pull/41

.github/workflows/release-package.yml
```diff
@@ -0,0 +1,29 @@
+name: Release Package
+
+on:
+  release:
+    types: [created]
+  push:
+    branches:
+      - main
```
gpad, line 8: @baldarn I think in this way every time we merge on main we publish a new version, I don't think it's correct.

gpad (approved):

PR #62, opened 2023-01-16
WIP:Store Offset

Yattol:
https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/7e9b3d8d7eed9cf45e6afa941b3f20feb8f02011/README.md#track-offset
Waiting for consumer to be developed. Tests still to be done


test/e2e/store_offset.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    await publisher.send(Buffer.from("hello"))
+    const consumer = await connection.declareConsumer(
+      { stream: testStreamName, consumerRef: "my consumer", offset: Offset.first() },
+      (message: Message) => messages.push(message.content)
+    )
+
+    await consumer.storeOffset(Offset.offset(1n).value!)
+
+    // TODO - change expect, similar to the credit one
+    await eventually(() => expect(messages).eql([Buffer.from("hello")]))
```
gpad, line 44: Is it possible to retrieve this information via API? Otherwise, I think we should merge the store and query  PR together.

src/consumer.ts
```diff
@@ -1,11 +1,34 @@
...
+    this.consumerRef = params.consumerRef || ""
+  }
 
   async close(): Promise<void> {
     throw new Error("Method not implemented.")
   }
+
+  public storeOffset(offsetValue: bigint): Promise<void> {
+    return this.connection.storeOffset({ stream: this.stream, reference: this.consumerRef, offsetValue })
+  }
```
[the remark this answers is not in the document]
albertobarrila: the C# implementation expose the call like we do 
https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/7e9b3d8d7eed9cf45e6afa941b3f20feb8f02011/README.md#track-offset 
I would stick to it.

test/e2e/store_offset.test.ts
albertobarrila: the api is our old friends /consumer that has already caused us some troubles, I think that we can use store offset to test query offset and vice versa when ready.

src/consumer.ts
```diff
@@ -1,11 +1,34 @@
...
+  public consumerId: number
+  public consumerRef: string
+
+  constructor(
+    readonly handle: ConsumerFunc,
+    params: {
+      connection: Connection
+      stream: string
+      consumerId: number
+      consumerRef: string
```
gpad, line 18: Are you sure it's mandatory for the conusmerRef??

test/e2e/store_offset.test.ts
gpad: Ok, but why don't try to replicate the sample of c# so store the offset while we are consuming message?

src/consumer.ts
albertobarrila: the store and retrieval of offset is based on the consumerRef so I'll keep it mandatory, otherwise the functionality need to check is the ref is present before act (and maybe throw an exception???)

test/e2e/store_offset.test.ts
albertobarrila: we can add the code as example like in the c# doc, but as a test a the moment we do not have the way to check it. after the merge we can improve the test

src/consumer.ts
gpad: I have checked the documentation in C# https://rabbitmq.github.io/rabbitmq-stream-dotnet-client/stable/htmlsingle/index.html#_creating_a_consumer and the consumerRef is not mandatory. It's used if you want to enable this feature ... https://rabbitmq.github.io/rabbitmq-stream-dotnet-client/stable/htmlsingle/index.html#consumer-offset-tracking So I suggest to implement it in the correct way. If you pass the consumerRef you can store the offset otherwise it should rise an exception.

src/consumer.ts
```diff
@@ -21,14 +21,15 @@ export class Consumer {
     this.connection = params.connection
     this.stream = params.stream
     this.consumerId = params.consumerId
-    this.consumerRef = params.consumerRef
+    this.consumerRef = params.consumerRef || undefined
```
gpad, line 24:
```suggestion
    this.consumerRef = params.consumerRef
```

gpad: Also on line `18 should be declared as optional

src/connection.ts
```diff
@@ -434,6 +443,7 @@ export interface DeclarePublisherParams {
 
 export interface DeclareConsumerParams {
   stream: string
+  consumerRef: string
```
gpad, line 446: This should be optional

test/e2e/credit.test.ts
```diff
@@ -36,8 +36,9 @@ describe.skip("credit management", () => {
       await publisher.send(m)
     }
 
-    await connection.declareConsumer({ stream: streamName, offset: Offset.first() }, (message: Message) =>
-      receivedMessages.push(message.content)
+    await connection.declareConsumer(
+      { stream: streamName, offset: Offset.first(), consumerRef: "my_consumer" },
+      (message: Message) => receivedMessages.push(message.content)
```
gpad, line 41: Same as here

test/e2e/declare_consumer.test.ts
```diff
@@ -48,18 +49,22 @@ describe("declare consumer", () => {
     await publisher.send(Buffer.from("world"))
     await publisher.send(Buffer.from("world"))
 
-    await connection.declareConsumer({ stream: streamName, offset: Offset.first() }, (message: Message) =>
-      messages.push(message.content)
+    await connection.declareConsumer(
+      { stream: streamName, offset: Offset.first(), consumerRef: "my_consumer" },
+      (message: Message) => messages.push(message.content)
```
gpad, line 54: same as here

test/e2e/declare_consumer.test.ts
```diff
@@ -48,18 +49,22 @@ describe("declare consumer", () => {
...
     const receivedMessages: Buffer[] = []
-    await connection.declareConsumer({ stream: streamName, offset: Offset.next() }, (message: Message) => {
-      receivedMessages.push(message.content)
-    })
+    await connection.declareConsumer(
+      { stream: streamName, offset: Offset.next(), consumerRef: "my_consumer" },
+      (message: Message) => {
+        receivedMessages.push(message.content)
+      }
+    )
```
gpad, line 67: same as here

test/e2e/declare_consumer.test.ts
```diff
@@ -72,8 +77,9 @@ describe("declare consumer", () => {
   it("declaring a consumer on a non-existing stream should raise an error", async () => {
     await expectToThrowAsync(
       () =>
-        connection.declareConsumer({ stream: nonExistingStreamName, offset: Offset.first() }, (message: Message) =>
-          console.log(message.content)
+        connection.declareConsumer(
+          { stream: nonExistingStreamName, offset: Offset.first(), consumerRef: "my_consumer" },
+          (message: Message) => console.log(message.content)
```
gpad, line 82: same as here

test/e2e/declare_consumer.test.ts
```diff
@@ -86,10 +92,13 @@ describe("declare consumer", () => {
...
-      messageProperties.push(message.messageProperties || {})
-      messages.push("JSON.stringify(message.properties?.correlationId) ||")
-    })
+    await connection.declareConsumer(
+      { stream: streamName, offset: Offset.first(), consumerRef: "my_consumer" },
+      (message: Message) => {
+        messageProperties.push(message.messageProperties || {})
+        messages.push("JSON.stringify(message.properties?.correlationId) ||")
+      }
+    )
```
gpad, line 101: also here

test/e2e/declare_consumer.test.ts
```diff
@@ -102,10 +111,13 @@ describe("declare consumer", () => {
...
-      messageApplicationProperties.push(message.applicationProperties || {})
-      messages.push("JSON.stringify(message.properties?.correlationId) ||")
-    })
+    await connection.declareConsumer(
+      { stream: streamName, offset: Offset.first(), consumerRef: "my_consumer" },
+      (message: Message) => {
+        messageApplicationProperties.push(message.applicationProperties || {})
+        messages.push("JSON.stringify(message.properties?.correlationId) ||")
+      }
+    )
```
gpad, line 120: also here

test/e2e/store_offset.test.ts
```diff
@@ -0,0 +1,39 @@
...
+describe("declare consumer", () => {
+  const rabbit = new Rabbit(username, password)
+  const testStreamName = "test-stream"
+  let connection: Connection
+
+  beforeEach(async () => {
+    await rabbit.createStream(testStreamName)
+
+    connection = await createConnection(username, password)
+  })
```
gpad, line 18:
```suggestion
  beforeEach(async () => {
    await rabbit.createStream(testStreamName)
    connection = await createConnection(username, password)
  })
```

test/e2e/store_offset.test.ts
```diff
@@ -0,0 +1,39 @@
...
+      { stream: testStreamName, consumerRef: "my consumer", offset: Offset.first() },
+      (message: Message) => messages.push(message.content)
+    )
+
+    await consumer.storeOffset(Offset.offset(1n).value!)
+
+    // TODO - change expect, similar to the credit one
+    await eventually(() => expect(messages).eql([Buffer.from("hello")]))
+  }).timeout(10000)
+})
```
gpad, line 39: Add a test that raise an exception if consumerRef is not set...

gpad (approved):

PR #68, opened 2023-01-30
integrate queryPublisherSequence in producer

Yattol:


test/unit/producer.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  beforeEach(async () => {
+    await rabbit.createStream(testStreamName)
+  })
+
+  afterEach(async () => {
+    await rabbit.deleteStream(testStreamName)
+  })
+
+  describe("Send: ", () => {
```
gpad, line 19: Why this double describe?

test/unit/producer.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    it("boot is true", async () => {
```
gpad, line 20: if try to read the test as sentence this name are not very explicative. I suggest a tet like:
```suggestion
    it("retrive publishing id from server when boot is true", async () => {
```

and in the test i suggest to firs send some message using one publishRef, and then using the same publishRef see that we publish with new id ...

same for boot false, we overwrite the stream.

gpad (commented):
I also suggest to add a test that when you omit publishRef and set boot to true we start emitting from 1

test/unit/producer.test.ts
```diff
@@ -0,0 +1,58 @@
+import { expect } from "chai"
+import { randomUUID } from "crypto"
+import { connect } from "../../src"
+import { Rabbit } from "../support/rabbit"
+
+describe("Producer", () => {
+  const rabbit = new Rabbit()
+  const testStreamName = "test-stream"
+  const publisherRef = randomUUID()
```
gpad, line 9: This publishRef remains the same for every test, I suggest to put it or in beforeEach or in every test

test/unit/producer.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  let publisherRef: string
+
+  beforeEach(async () => {
+    publisherRef = randomUUID()
+    await rabbit.createStream(testStreamName)
+  })
+
+  afterEach(async () => {
+    await rabbit.deleteStream(testStreamName)
+  })
```
gpad, line 18:
```suggestion
  afterEach(() => rabbit.deleteStream(testStreamName))
```

gpad (dismissed):
a little comment form my OCD but I like it

PR #74, opened 2023-02-08
adding license

baldarn:
@gpad please check if the `ragione sociale` is correct
`coders51 S.R.L.`, because I recall something strange that @keown told me :)


gpad:
@baldarn which type fo license is it? I'm not an expert. Is it a MIT license?

gpad (approved):

PR #75, opened 2023-02-09
Credit

l4mby:
Handle credit response from server

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -10,7 +10,7 @@ import { OpenResponse } from "./responses/open_response"
 import { SaslHandshakeResponse } from "./responses/sasl_handshake_response"
 import { SaslAuthenticateResponse } from "./responses/sasl_authenticate_response"
 import { Response } from "./responses/response"
-import { MetadataUpdateListener, ResponseDecoder } from "./response_decoder"
+import { DeliverListener, MetadataUpdateListener, CreditListener, ResponseDecoder } from "./response_decoder"
 import { createConsoleLog, removeFrom } from "./util"
 import { WaitingResponse } from "./waiting_response"
 import { TuneResponse } from "./responses/tune_response"
@@ -32,6 +32,7 @@ import { SubscribeResponse } from "./responses/subscribe_response"
 import { Offset, SubscribeRequest } from "./requests/subscribe_request"
 import { Consumer, ConsumerFunc } from "./consumer"
 import { DeliverResponse } from "./responses/deliver_response"
+import { CreditResponse } from "./responses/credit_response"
 
 export class Connection {
   private readonly socket = new Socket()
@@ -87,8 +88,21 @@ export class Connection {
     })
   }
 
-  public on(_event: "metadata_update", listener: MetadataUpdateListener) {
-    this.decoder.on("metadata_update", listener)
+  public on(
+    _event: "metadata_update" | "credit_response" | "deliver",
+    listener: MetadataUpdateListener | CreditListener | DeliverListener
+  ) {
+    switch (_event) {
+      case "metadata_update":
+        this.decoder.on("metadata_update", listener)
+        break
+      case "credit_response":
+        this.decoder.on("credit_response", listener)
+        break
+      case "deliver":
+        this.decoder.on("deliver", listener)
+        break
+    }
   }
 
   public async close(
@@ -329,7 +343,10 @@ export class Connection {
   }
 
   private registerListeners(listeners?: ListenersParams) {
-    if (listeners) this.decoder.on("metadata_update", listeners.metadata_update)
+    if (listeners) {
+      this.on("metadata_update", listeners.metadata_update)
+      this.on("credit_response", listeners.credit)
+    }
   }
 
   private registerDelivers() {
@@ -340,7 +357,10 @@ export class Connection {
   }
 }
 
-type ListenersParams = Record<"metadata_update", MetadataUpdateListener>
+type ListenersParams = {
+  metadata_update: MetadataUpdateListener
+  credit: CreditListener
+}
 
 export interface ConnectionParams {
   hostname: string
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -32,7 +32,6 @@ import { SubscribeResponse } from "./responses/subscribe_response"
 import { Offset, SubscribeRequest } from "./requests/subscribe_request"
 import { Consumer, ConsumerFunc } from "./consumer"
 import { DeliverResponse } from "./responses/deliver_response"
-import { CreditResponse } from "./responses/credit_response"
 import { CreditRequest, CreditRequestParams } from "./requests/credit_request"
 
 export class Connection {
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -10,7 +10,7 @@ import { OpenResponse } from "./responses/open_response"
 import { SaslHandshakeResponse } from "./responses/sasl_handshake_response"
 import { SaslAuthenticateResponse } from "./responses/sasl_authenticate_response"
 import { Response } from "./responses/response"
-import { DeliverListener, MetadataUpdateListener, CreditListener, ResponseDecoder } from "./response_decoder"
+import { MetadataUpdateListener, CreditListener, ResponseDecoder } from "./response_decoder"
 import { createConsoleLog, removeFrom } from "./util"
 import { WaitingResponse } from "./waiting_response"
 import { TuneResponse } from "./responses/tune_response"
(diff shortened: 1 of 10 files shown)


src/connection.ts
```diff
@@ -87,8 +88,21 @@ export class Connection {
     })
   }
 
-  public on(_event: "metadata_update", listener: MetadataUpdateListener) {
-    this.decoder.on("metadata_update", listener)
+  public on(
+    _event: "metadata_update" | "credit_response" | "deliver",
```
gpad, line 92: the `_` is only for arguments not used so rename ti in `event`

gpad: Why do you have also exposed also the `deliver` event? I suggest to keep it "private"

src/response_decoder.ts
```diff
@@ -84,6 +99,11 @@ function decodeResponse(
     }
     return { size, key, version, metadataInfo } as RawMetadataUpdateResponse
   }
+  if (key === CreditResponse.key) {
+    const responseCodeCredit = dataResponse.readUInt16()
+    const subscriptionId = dataResponse.readUInt8()
+    return { size, key, version, responseCode: responseCodeCredit, subscriptionId: subscriptionId } as RawCreditResponse
```
gpad, line 105: I suggest a different management like:
```
    const response: RawDeliverResponse = {
      size,
      key: key as DeliverResponse["key"],
      version,
      subscriptionId,
      messages,
    }
return response
```
if you want to know  why @ggllcu will explain it.

src/response_decoder.ts
```diff
@@ -233,29 +253,65 @@ export class BufferDataReader implements DataReader {
...
 
 function isHeartbeatResponse(
-  params: RawResponse | RawTuneResponse | RawHeartbeatResponse | RawMetadataUpdateResponse | RawDeliverResponse
+  params:
+    | RawResponse
+    | RawTuneResponse
+    | RawHeartbeatResponse
+    | RawMetadataUpdateResponse
+    | RawDeliverResponse
+    | RawCreditResponse
```
gpad, line 274: I suggest to create an alias so we can change it easily

src/connection.ts
```diff
@@ -234,6 +248,10 @@ export class Connection {
     return res
   }
 
+  public async askForCredit(params: CreditRequestParams): Promise<void> {
+    return await this.send(new CreditRequest({ ...params }))
+  }
```
gpad, line 253: I don't think we have to expose this functionality to the final user, I think we have to embed it in the consumer, check the .NET client to see an example

l4mby: Ok, i saw that the on the .NET client the credit is used on the RawConsumer, should we replicate this the same way in our consumer? Can I use the RawConsumer as a reference?

src/connection.ts
```diff
@@ -87,8 +88,8 @@ export class Connection {
     })
   }
 
-  public on(_event: "metadata_update", listener: MetadataUpdateListener) {
-    this.decoder.on("metadata_update", listener)
+  public on(event: "metadata_update" | "credit_response", listener: MetadataUpdateListener | CreditListener) {
+    this.decoder.on(event, listener)
```
gpad, line 92: I suggest to don't expose the credit_response as event in connection

src/connection.ts
```diff
@@ -329,18 +334,25 @@ export class Connection {
   }
 
   private registerListeners(listeners?: ListenersParams) {
-    if (listeners) this.decoder.on("metadata_update", listeners.metadata_update)
+    if (listeners) {
+      this.on("metadata_update", listeners.metadata_update)
+      this.on("credit_response", listeners.credit)
```
gpad, line 339: same as above

src/connection.ts
```diff
@@ -234,6 +235,10 @@ export class Connection {
     return res
   }
 
+  public askForCredit(params: CreditRequestParams): Promise<void> {
+    return this.send(new CreditRequest({ ...params }))
+  }
+
```
gpad, line 241: I suggest keeping this function private

src/connection.ts
```diff
@@ -329,18 +334,25 @@ export class Connection {
...
+      await this.askForCredit({ credit: 1, subscriptionId: response.subscriptionId })
       response.messages.map((x) => this.consumers[response.subscriptionId].handle(x))
     })
   }
 }
 
-type ListenersParams = Record<"metadata_update", MetadataUpdateListener>
+type ListenersParams = {
+  metadata_update: MetadataUpdateListener
+  credit: CreditListener
```
gpad, line 354: as above

test/e2e/credit.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      subscriptionId: 1,
+      stream: streamName,
+      offset: Offset.next(),
+      credit: 1,
+    })
+
+    await connection.askForCredit({ subscriptionId: 100, credit: 1 })
+
+    await eventually(async () => expect(creditResponses.length).greaterThanOrEqual(1), 10000)
+  }).timeout(10000)
```
gpad, line 65: If you move `askForCredit` as private I suggest to remove these test

test/e2e/credit.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    for (const m of messages) {
+      await publisher.send(m)
+    }
+
+    await connection.declareConsumer({ stream: streamName, offset: Offset.first() }, (message: Message) =>
+      receivedMessages.push(message.content)
+    )
+
+    await eventually(async () => expect(await rabbit.returnSingleConsumerCredits()).eql(10), 5000)
+    await eventually(() => expect(receivedMessages).eql(messages), 1500)
```
gpad, line 82: I don't think the timeout at the enc is necessary ...
```suggestion
    await eventually(async () => expect(await rabbit.returnSingleConsumerCredits()).eql(10))
    await eventually(() => expect(receivedMessages).eql(messages))
```

test/e2e/metadata_update.test.ts
```diff
@@ -20,7 +20,10 @@ describe("update the metadata from the server", () => {
       vhost: "/",
       frameMax: 0, // not used
       heartbeat: 0, // not used
-      listeners: { metadata_update: (data) => metadataUpdateResponses.push(data) },
+      listeners: {
+        metadata_update: (data) => metadataUpdateResponses.push(data),
+        credit: (_data) => console.error("Subscribe server error"),
```
gpad, line 25: same as above

test/e2e/metadata_update.test.ts
```diff
@@ -40,7 +43,7 @@ describe("update the metadata from the server", () => {
     let called = 0
     await rabbit.createStream(stream)
     await connection.declarePublisher({ stream, publisherRef: "my publisher" })
-    connection.on("metadata_update", (_) => called++)
+    connection.on("metadata_update", (_data: MetadataUpdateResponse) => called++)
```
gpad, line 46: ???

test/support/rabbit.ts
```diff
@@ -117,6 +133,25 @@ export class Rabbit {
...
+    const resp = await got.get<RabbitConnectionDetails[]>(
+      `http://localhost:15672/api/stream/connections/%2F/${connectionName}/consumers`,
+      {
+        username: "rabbit",
+        password: "rabbit",
+        responseType: "json",
+      }
+    )
+    return resp.body[0].credits
+  }
```
gpad, line 153: I suggest returning an array for every `consumerChannelDetails`, and then in the test, you get the first. In this way, this method is more generic. Clearly, you have also to change the name to `returnConsumerCredits` or, better return all the information you have and select the first item and the credit in the test.

albertobarrila (commented):
LGTM

albertobarrila (approved):
LGTM

src/response_decoder.ts
```diff
@@ -84,10 +88,17 @@ function decodeResponse(
     }
     return { size, key, version, metadataInfo } as RawMetadataUpdateResponse
   }
+
+  if (key === CreditResponse.key) {
+    const responseCode = dataResponse.readUInt16()
+    const subscriptionId = dataResponse.readUInt8()
+    return { size, key, version, responseCode, subscriptionId } as RawCreditResponse
```
gpad, line 95: I suggest using a different way:
```suggestion
   const ret: RawCreditResponse = { size, key, version, responseCode, subscriptionId }
    return  ret
```
with the `as` if we change a field or make some refactoring, the compiler doesn't raise an error

src/response_decoder.ts
```diff
@@ -288,13 +295,19 @@ export class ResponseDecoder {
       } else if (isDeliverResponse(response)) {
         this.emitter.emit("deliver", new DeliverResponse(response))
         this.logger.debug(`deliver received from the server: ${inspect(response)}`)
+      } else if (isCreditResponse(response)) {
+        this.emitter.emit("credit_response", new CreditResponse(response))
+        this.logger.debug(`credit received from the server: ${inspect(response)}`)
```
gpad, line 300: I suggest making the log and the raise the event, this event is synchronous so in the log, if you switch it it's more readeble

test/e2e/credit.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    await rabbit.createStream(streamName)
+  })
+
+  afterEach(async () => {
+    await connection.close()
+    await rabbit.deleteStream(streamName)
+  })
+
+  // This test can only run locally, the HTTP API gives different results in GitHub CI (https://coders51.slack.com/archives/C03E263HH38/p1681374600592829)
+  it.skip(`the number of credit remain stable after have consumed some messages`, async () => {
```
gpad, line 35: This test fail because (I think) there is a bug in the consumer see the pr #86. But it's fine to skip it for now. Can you create a card to fix this test in the feature?

albertobarrila (approved):
all notes have been fixed, so for me is ok

PR #76, opened 2023-03-30
65 consuming

ggllcu:
declaring a consumer on an existing stream and handling the messages


src/response_decoder.ts
```diff
@@ -33,17 +36,84 @@ import { SubscribeResponse } from "./responses/subscribe_response"
...
+          switch (type) {
+            case FormatCode.Vbin8:
+              const length = dataResponse.readUInt8()
+              messages.push(dataResponse.readToBuffer(length))
+          }
+          break
+        default:
+          break
+      }
+    }
```
gpad, line 107: Extract this code in a function

src/responses/deliver_response.ts
```diff
@@ -0,0 +1,51 @@
...
+    return this.response.subscriptionId
+  }
+
+  get messages(): Message[] {
+    return this.response.messages.map((x) => {
+      return {
+        content: x,
+      }
+    })
+  }
```
gpad, line 50:
```suggestion
  get messages(): Message[] {
    return this.response.messages.map((x) => ({ content: x}))
  }
```

src/response_decoder.ts
```diff
@@ -76,6 +153,12 @@ class BufferDataReader implements DataReader {
     return ret
   }
 
+  readToBuffer(size: number): Buffer {
```
gpad, line 156:
```suggestion
  readBuffer(size: number): Buffer {
```
Like readInt8, readInt16, readBuffer read a buffer

test/e2e/declare_consumer.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  it("declaring a consumer on an existing stream - the consumer should handle the message", async () => {
+    const connection = await connect({
+      hostname: "localhost",
+      port: 5552,
+      username: "rabbit",
+      password: "rabbit",
+      vhost: "/",
+      frameMax: 0,
+      heartbeat: 0,
+    })
```
gpad, line 30: I suggest to move the connection in beforeEach (and its close in afterEach) so if an test fail the connection is closed

gpad (approved):

src/consumer.ts
```diff
@@ -0,0 +1,11 @@
+import { Message } from "./producer"
+
+export type ConsumerFunc = (message: Message) => void
+
+export class Consumer {
+  constructor(readonly handle: ConsumerFunc) {}
+
+  async close(): Promise<void> {
+    throw new Error("Method not implemented.")
+  }
```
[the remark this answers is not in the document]
ggllcu: I think it was put there at the beginning, but I don't know if it makes sense. There is no "close" in the producer. I asked Alberto and according to him at the moment it could even be removed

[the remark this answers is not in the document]
gpad: @albertobarrila @ggllcu I suggest verifying the list of commands and seeing if there is a command to "stop consuming". If there, I suggest adding a card. if There isn't, I suggest removing this command in the next PR. In the meanwhile, I merge this PR.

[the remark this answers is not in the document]
gpad: @albertobarrila @ggllcu I have created this Issue https://github.com/coders51/rabbitmq-stream-js-client/issues/77

PR #84, opened 2023-04-14
78 - Add credential parameters to rabbit object constructor

l4mby:


test/e2e/basic_publish.test.ts
```diff
@@ -7,7 +7,7 @@ import * as ampq from "amqplib"
 import { MessageProperties } from "../../src/producer"
 
 describe("publish a message", () => {
-  const rabbit = new Rabbit()
+  const rabbit = new Rabbit("rabbit", "rabbit")
```
gpad, line 10:
```suggestion

  import {username, password} from "../support/util"

// ...
  const rabbit = new Rabbit(username, password)
  
  // also in connection 
  
  connection = await connect({
      hostname: "localhost",
      port: 5552,
      username,
      password,
      vhost: "/",
      frameMax: 0, // not used
      heartbeat: 0,
})
```

in util:

```typescript
export const username = process.ENV.RABBIT_USER
export const password = process.ENV.RABBIT_PASSWORD
```

And then change it also in the `/.github/workflows/main.yml` to setup a different username and password I suggest to set different between username and password something like `test-user` and `test-passowrd`

example/index.js
```diff
@@ -2,15 +2,18 @@ const rabbit = require("rabbitmq-stream-js-client")
...
 const { randomUUID } = require("crypto")
 
+const rabbitUser = process.env.RABBITMQ_USER || "rabbit"
+const rabbitPassword = process.env.RABBITMQ_PASSWORD || "rabbit"
+
 async function main() {
   const streamName = `example-${randomUUID()}`
   console.log(`Create stream ${streamName}`)
-
+  console.log("AAAA", process.env.RABBITMQ_USER, process.env.RABBITMQ_PASSWORD)
```
gpad, line 11: ???

gpad (approved):

PR #91, opened 2023-05-16
DeletePublisher Command

ablease:
Hello! This PR adds the DeletePublisher Command

https://github.com/rabbitmq/rabbitmq-server/blob/1ee8454129b7769fda6dfc1a528e886b290aa258/deps/rabbitmq_stream/docs/PROTOCOL.adoc#deletepublisher

connection.ts can handle sending DeletePublisher requests, and parsing DeletePublisherResponse.

I also exposed the publishingId on the producer class, since publishingId is required in DeletePublisherRequest. 

Tests passing with `npm run test`

### DIFF
--- a/src/producer.ts
+++ b/src/producer.ts
@@ -101,7 +101,17 @@ export class Producer {
     )
   }
 
+<<<<<<< HEAD
   getLastPublishingId(): Promise<bigint> {
+||||||| parent of dc0ccf1 (DeletePublisher Command)
+  getLastPublishingId() {
+=======
+  getPublisherId() {
+    return this.publisherId
+  }
+
+  getLastPublishingId() {
+>>>>>>> dc0ccf1 (DeletePublisher Command)
     return this.connection.queryPublisherSequence({ stream: this.stream, publisherRef: this.publisherRef })
   }
 
--- a/src/producer.ts
+++ b/src/producer.ts
@@ -34,7 +34,7 @@ type PublishConfirmCallback = (err: number | null, publishingIds: bigint[]) => v
 export class Producer {
   private connection: Connection
   private stream: string
-  private publisherId: number
+  readonly publisherId: number
   private publisherRef: string
   private boot: boolean
   private publishingId: bigint
@@ -101,17 +101,7 @@ export class Producer {
     )
   }
 
-<<<<<<< HEAD
   getLastPublishingId(): Promise<bigint> {
-||||||| parent of dc0ccf1 (DeletePublisher Command)
-  getLastPublishingId() {
-=======
-  getPublisherId() {
-    return this.publisherId
-  }
-
-  getLastPublishingId() {
->>>>>>> dc0ccf1 (DeletePublisher Command)
     return this.connection.queryPublisherSequence({ stream: this.stream, publisherRef: this.publisherRef })
   }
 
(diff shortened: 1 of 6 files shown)


test/unit/delete_publisher.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  it("can delete a publisher", async () => {
+    const connection = await connect({
+      hostname: "localhost",
+      port: 5552,
+      username,
+      password,
+      vhost: "/",
+      frameMax: 0,
+      heartbeat: 0,
+    })
```
gpad, line 28: I suggest to use `createConnection` function from test/support/fake_data.ts

test/unit/delete_publisher.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    const publisher = await connection.declarePublisher({ stream: testStreamName, publisherRef })
+    await publisher.send(Buffer.from(`test${randomUUID()}`))
+
+    const publisherId = publisher.getPublisherId()
+
+    const deletePublisher = await connection.deletePublisher(Number(publisherId))
+    expect(deletePublisher).eql(true)
+
+    await connection.close()
```
gpad, line 37: When write a test please keep the AAA style:http://wiki.c2.com/?ArrangeActAssert 
```suggestion
    const publisher = await connection.declarePublisher({ stream: testStreamName, publisherRef })
    await publisher.send(Buffer.from(`test${randomUUID()}`))

    const publisherId = publisher.getPublisherId()

    const deletePublisher = await connection.deletePublisher(Number(publisherId))
    expect(deletePublisher).eql(true)
    await connection.close()
```

test/unit/delete_publisher.test.ts
gpad: I suggest to open and close the connection in the beforeEach and closeEach because closing the connection as last statement of test, if the test fails, can cause the connection remain pending

src/producer.ts
```diff
@@ -99,6 +99,10 @@ export class Producer {
     )
   }
 
+  getPublisherId() {
+    return this.publisherId
+  }
+
```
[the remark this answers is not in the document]
ablease: I was looking for a way for the tests to have access to the publisherId for a publisher. The Consumer does this by [exposing consumerId as a readonly property](https://github.com/coders51/rabbitmq-stream-js-client/blob/a72b2cc3fabe0e7964cb88f5463c41f9cd4f8a58/src/consumer.ts#L6). Would that be acceptable? I have made that change in this commit https://github.com/coders51/rabbitmq-stream-js-client/blob/7cbc2d08f70627cf6a7270380a5f4e9019919369/src/producer.ts#L36

test/unit/delete_publisher.test.ts
ablease: Thanks for pointing that out. I've made use of the createConnection in fake_data.ts, and open and closed teh connection in before/aftereach in this commit https://github.com/coders51/rabbitmq-stream-js-client/commit/7cbc2d08f70627cf6a7270380a5f4e9019919369

test/unit/delete_publisher.test.ts
ablease: Resloved in this commit https://github.com/coders51/rabbitmq-stream-js-client/commit/7cbc2d08f70627cf6a7270380a5f4e9019919369

ablease:
Hi @gpad looking for review here. Thanks!

gpad (approved):
Hi @ablease, first of all, thanks for your contribution. It's fine for me!!!

PR #92, opened 2023-05-22
StreamStats Command

ablease:
```
StreamStatsRequest => Key Version CorrelationId Stream
  Key => uint16 // 0x001c
  Version => uint16
  CorrelationId => uint32
  Stream => string

StreamStatsResponse => Key Version CorrelationId ResponseCode Stats
  Key => uint16 // 0x801c
  Version => uint16
  CorrelationId => uint32
  ResponseCode => uint16
  Stats => [Statistic]
  Statistic => Key Value
  Key => string
  Value => int64
```

Add's streamStatsRequest function to connection. This returns an error if one occurs, or a map of stream stats.

Given the values for `stats.committed_chunk_id`, `stats.first_chunk_id` and `stats.last_chunk_id` can vary, I instead asserted on the structure and data types of the payload, rather than the actual values. 

I can't think of a way of asserting on the values without mocking rabbitmq-server entirely. Let me know if you can think of a way!


test/unit/stream_stats.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    const publisher = await connection.declarePublisher({ stream: testStreamName, publisherRef })
+    for (let i = 0; i < 5; i++) {
+      await publisher.send(Buffer.from(`test${randomUUID()}`))
+    }
+
+    const stats = await connection.streamStatsRequest(testStreamName)
+
+    expect(stats.committed_chunk_id).to.be.a("BigInt")
+    expect(stats.first_chunk_id).to.be.a("BigInt")
+    expect(stats.last_chunk_id).to.be.a("BigInt")
```
gpad, line 35: Do you think it's possible to make some assertion about the "validity" of data, something like:

I have sent 5 messages committed_chunk_id should be 5?

WDYT?

src/responses/stream_stats_response.ts
```diff
@@ -0,0 +1,19 @@
+import { AbstractResponse } from "./abstract_response"
+import { RawResponse } from "./raw_response"
+
+export class StreamStatsResponse extends AbstractResponse {
+  static key = 0x801c
+  readonly statistics: Record<string, bigint> = {}
```
gpad, line 6: Is it possible to define the keys `committed_chunk_id` like for the MessageProperties (reployTo, correlationId, etc ..) ?
If it's possible, I suggest it to translate in camelcase committedChunkId

test/unit/stream_stats.test.ts
ablease: Hi @gpad. I did think about that. In my testing the value of `commited_chunk_id` can vary. For example sending 5 messages, it can be 1, 2, or sometimes 3. 

When I tested with just publishing 1 message in the test, the `commited_chunk_id` varied between -1 and 0.

This behaviour is expected, for example see this doc in the java client https://rabbitmq.github.io/rabbitmq-stream-java-client/snapshot/api/com/rabbitmq/stream/StreamStats.html#committedChunkId()

> The value can be stale as soon as the application reads it though, as the committed chunk ID for a stream that is published to changes all the time.

src/responses/stream_stats_response.ts
ablease: Yes I agree. I made this change to accomodate that. https://github.com/coders51/rabbitmq-stream-js-client/pull/92/commits/ead69e5a54f30f52cbe8cadd59510ba624a75ee4

gpad (dismissed):

ablease:
Hi @gpad  I had to merge in some updates form main branch, and this needs review again. Thank you!

gpad (approved):

Issue #83, opened 2023-04-12
Consumer should read message AmqpValue


ggllcu:
This issue is a continuation of #65

Issue #82, opened 2023-04-12
Consumer should read message header


ggllcu:
This issue is a continuation of #65

PR #94, opened 2023-06-01
82 - Decoding of messageHeader and amqpValue

l4mby:
The feature is only for the decoding of these message parts in accordance with protocol amqp 1.0
The tests were done using a binary file containing the message to decode with messageHeader and amqpValue
Ref #82 #83


src/producer.ts
```diff
@@ -94,6 +105,7 @@ export class Producer {
             message: {
               content: message,
               messageProperties: opts.messageProperties,
+              messageHeader: opts.messageHeader,
```
gpad, line 108: Could the header be sent? are you sure?

src/response_decoder.ts
```diff
@@ -203,14 +206,49 @@ function decodeMessage(dataResponse: DataReader, offset: bigint): Message {
...
+        break
+      default:
+        throw new Error(`Not supported format code ${formatCode}`)
+    }
+  }
+
+  return { content, messageProperties, messageHeader, applicationProperties, amqpValue, offset }
+}
+
+export function decodeMessageTesting(dataResponse: DataReader, length: number): Message {
```
gpad, line 222: What is this function? If it's used in test should be declared in test

src/response_decoder.ts
```diff
@@ -183,12 +184,14 @@ function decodeDeliverResponse(dataResponse: DataReader, logger: Logger): Delive
 
 const EmptyBuffer = Buffer.from("")
 
-function decodeMessage(dataResponse: DataReader, offset: bigint): Message {
+export function decodeMessage(dataResponse: DataReader, offset: bigint): Message {
```
gpad, line 187: why export?

src/response_decoder.ts
```diff
@@ -238,6 +276,23 @@ function decodeMessageProperties(dataResponse: DataReader) {
   return Properties.parse(dataResponse, propertiesLength as number)
 }
 
+export function decodeMessageHeader(dataResponse: DataReader) {
```
gpad, line 279: why export?

test/support/util.ts
```diff
@@ -89,5 +89,13 @@ export async function createClassicConsumer(
 
   return { conn, ch }
 }
+
+export async function createClassicPublisher(): Promise<{ conn: ampq.Connection; ch: ampq.Channel }> {
+  const conn = await ampq.connect(`amqp://${username}:${password}@localhost`)
+  const ch = await conn.createChannel()
+
+  return { conn, ch }
+}
```
gpad, line 98:
```suggestion
export async function createClassicPublisher(): Promise<{ conn: ampq.Connection; ch: ampq.Channel }> {
  const conn = await ampq.connect(`amqp://${username}:${password}@localhost`)
  const ch = await conn.createChannel()
  return { conn, ch }
}
```

src/producer.ts
l4mby: Sorry my bad, I added this before I knew it shouldn't be sent. I will fix this

albertobarrila (approved):
all previous comments has been resolved so looks good to me

gpad (approved):

PR #97, opened 2023-06-02
RouteQuery/Response

### LINKED ISSUE #95: Route Command (SuperStream)
> A super stream is made of several individual streams, so it can be considered a logical entity rather than an actual physical entity. 

https://github.com/rabbitmq/rabbitmq-server/blob/1ee8454129b7769fda6dfc1a528e886b290aa258/deps/rabbitmq_stream/docs/PROTOCOL.adoc#route

RouteQuery, is initiated by the client, and the server returns an array of stream names associated with a Superstream.

```
RouteQuery => Key Version CorrelationId RoutingKey SuperStream
  Key => uint16 // 0x0018
  Version => uint16
  CorrelationId => uint32
  RoutingKey => string
  SuperStream => string

RouteResponse => Key Version CorrelationId ResponseCode [Stream]
  Key => uint16 // 0x8018
  Version => uint16
  CorrelationId => uint32
  ResponseCode => uint16
  Stream => string
  ```

ablease:
closes #95 
RouteQuery returns a stream or set of streams for a given routing key and superStream.

In the test I'm using exec to create a superstream with the rabbitmqctl cli. The result of that command is something like this. I'll have to figure out a way to do something like this in CI.
Screenshot 2023-06-02 at 17 01 00

This allows us to make successful RouteRequests for a given routingKey (0,1,2 are the default routing keys for the subsequent partition streams).


albertobarrila (approved):
lgtm

PR #107, opened 2023-11-16
106 add logger configuration

albertobarrila:


gpad (commented):
@albertobarrila What I suggested was to define a new Logger interface (not related to winston) that you can pass optionally to connection. If present, the user will use it, otherwise it will use a NullLogger. In this way we can remove the winston dependency from the library and leave to the caller manage and configure the logger. WDYT?

albertobarrila:
sounds right what about having 

```typescript
export interface Logger {
  debug(message: string): void
  info(message: string): void
  error(message: string): void
  warn(message: string): void
}

export class NullLogger implements Logger {
  debug(_message: string): void {
    // do nothing
  }
  info(_message: string): void {
    // do nothing
  }
  error(_message: string): void {
    // do nothing
  }
  warn(_message: string): void {
    // do nothing
  }
}
```
and having the 

```typescript 
export class Connection {

//...

static connect(params: ConnectionParams, logger?: Logger): Promise<Connection> {
    return new Connection(logger ?? new NullLogger()).start(params)
  }

//...

}
```
you mean something like this?

PR #110, opened 2023-11-30
106-add-logger-configuration

l4mby:


gpad (commented):
I think you could at least move winston from dependencies to devDependencies...
Is it possible to completely remove it?

test/e2e/declare_consumer.test.ts
```diff
@@ -72,7 +72,7 @@ describe("declare consumer", () => {
     await eventually(() => expect(messages).eql([Buffer.from("hello"), Buffer.from("world"), Buffer.from("world")]))
   }).timeout(10000)
 
-  it(`consume a lot of messages`, async () => {
+  it.skip(`consume a lot of messages`, async () => {
```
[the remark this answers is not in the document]
albertobarrila: yes it's fixxed in the https://github.com/coders51/rabbitmq-stream-js-client/pull/112

src/amqp10/messageHeader.ts
```diff
@@ -12,7 +12,7 @@ export class Header {
           const formatCode = dataResponse.readUInt8()
           dataResponse.rewind(1)
           const decodedBoolean = decodeFormatCode(dataResponse, formatCode)
-          if (!decodedBoolean) throw new Error(`invalid formatCode %#02x: ${formatCode}`)
+          if (decodedBoolean === undefined) throw new Error(`invalid formatCode %#02x: ${formatCode}`)
```
gpad, line 15: Why? I strongly suggest to never check directly `undefined` or `null` I think the previous code is much better

src/response_decoder.ts
```diff
@@ -264,12 +264,12 @@ function decodeMessageHeader(dataResponse: DataReader) {
...
 
-  const nextType = dataResponse.readInt8()
-  decodeFormatCode(dataResponse, nextType)
+  // next, the composite type is encoded as an AMQP uint8
+  dataResponse.readUInt64()
 
   const formatCode = dataResponse.readUInt8()
   const headerLength = decodeFormatCode(dataResponse, formatCode)
-  if (!headerLength) throw new Error(`invalid formatCode %#02x: ${formatCode}`)
+  if (headerLength === undefined) throw new Error(`invalid formatCode %#02x: ${formatCode}`)
```
gpad, line 272: again

test/support/fake_data.ts
```diff
@@ -42,5 +42,7 @@ export async function createConnection(username: string, password: string, liste
     frameMax: 0, // not used
     heartbeat: 0,
     listeners: listeners,
-  })
+  }
+
+  return await connect(params)
```
gpad, line 47:
```typescript
export function createConnection(username: string, password: string, listeners?: ListenersParams) {
  return connect({
    hostname: "localhost",
    port: 5552,
    username,
    frameMax: 0, 
    heartbeat: 0,
    listeners: listeners,
  })
}

```

src/response_decoder.ts
```diff
@@ -264,12 +264,12 @@ function decodeMessageHeader(dataResponse: DataReader) {
     throw new Error(`invalid composite header %#02x: ${type}`)
   }
 
-  const nextType = dataResponse.readInt8()
-  decodeFormatCode(dataResponse, nextType)
+  // next, the composite type is encoded as an AMQP uint8
+  dataResponse.readUInt64()
```
gpad, line 268: Are you sure about it?

src/amqp10/messageHeader.ts
l4mby: The problem is that in the previous code even if the decodedBoolean is false it throws an exception, but false is correct

src/response_decoder.ts
l4mby: Same as above, if headerLength is 0 it throws the exception, but 0 can be a correct value

src/response_decoder.ts
l4mby: This is also done in the dotnet client 
https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/main/RabbitMQ.Stream.Client/AMQP/AmqpWireFormattingRead.cs#L389

gpad (approved):

PR #113, opened 2023-12-01
55 ability to consume a batch of messages without compression

tarzacodes:
See the issue. We can now read from the batch subentries


tarzacodes:
The pipeline is red because of the usual rabbitMQ plugins in pipeline issue. The tests are green if you run them locally

albertobarrila (approved):
we have fixxed the "wrong" things with @gpad

PR #117, opened 2023-12-06
56-ability-to-consume-a-batch-of-messages-with-gzip-compression

l4mby:


albertobarrila (approved):
LGTM, it's just the test the behaviour is already in main

PR #122, opened 2023-12-12
121 support tls

albertobarrila:


src/connection.ts
```diff
@@ -89,29 +96,31 @@ export class Connection {
...
 
   static connect(params: ConnectionParams, logger?: Logger): Promise<Connection> {
-    return new Connection(logger ?? new NullLogger()).start(params)
+    return new Connection(logger ?? new NullLogger(), params).start()
   }
 
-  public start(params: ConnectionParams): Promise<Connection> {
-    this.registerListeners(params.listeners)
+  public start(): Promise<Connection> {
+    this.registerListeners(this.params.listeners)
```
tarzacodes, line 103: Would have been easier to simply add a line `const params = this.params`, makes everything else less verbose

src/response_decoder.ts
```diff
@@ -314,15 +314,15 @@ function decodeMessageHeader(dataResponse: DataReader) {
 
   const formatCode = dataResponse.readUInt8()
   const headerLength = decodeFormatCode(dataResponse, formatCode)
-  if (!headerLength) throw new Error(`invalid formatCode %#02x: ${formatCode}`)
+  //if (!headerLength) throw new Error(`invalid formatCode %#02x: ${formatCode}`)
```
tarzacodes, line 318: Why was this commented out?

albertobarrila: I've removed it it was dead code

tarzacodes (approved):

src/connection.ts
```diff
@@ -482,6 +493,12 @@ export type ListenersParams = {
   publish_error?: PublishErrorListener
 }
 
+export interface SSLConnectionParams {
+  key: string
+  cert: string
+  ca?: string
+}
```
ThomasShih, line 500: These Params probably should be all optional and another key value pair for isEnabled? there may a use case where we would like to use TLS but we don't need to pass a key or cert.

PR #124, opened 2023-12-13
Adds Metadata query

tarzacodes:


src/connection.ts
```diff
@@ -157,6 +159,18 @@ export class Connection {
     return new Promise((res, _rej) => this.socket.end(() => res()))
   }
 
+  public async queryMetadata(params: QueryMetadataParams): Promise<Record<string, StreamMetadata>> {
```
tarzacodes, line 162: Return a list instead of a Record

albertobarrila (approved):
LGTM

PR #125, opened 2023-12-14
19 - Adds frame max negotiation

l4mby:


src/connection.ts
```diff
@@ -463,6 +468,13 @@ export class Connection {
       response.messages.map((x) => consumer.handle(x))
     })
   }
+
+  private calculateFrameMaxSizeFrom(tuneResponseFrameMax: number) {
+    if (this.frameMax === DEFAULT_UNLIMITED_FRAME_SIZE && tuneResponseFrameMax === DEFAULT_UNLIMITED_FRAME_SIZE) {
+      return 0
+    }
+    return Math.max(1, Math.min(this.frameMax, tuneResponseFrameMax))
```
icappello, line 476: If I'm reading this correctly, if `tuneResponseFrameMax = 0` and `this.frameMax = 100` this expression returns 1 instead of 100.

I suggest something like this in essence:

```suggestion
    if (this.frameMax === DEFAULT_UNLIMITED_FRAME_SIZE) return tuneResponseFrameMax
    if (tuneResponseFrameMax  === DEFAULT_UNLIMITED_FRAME_SIZE) return this.frameMax
return Math.min(this.frameMax, tuneResponseFrameMax)

```

icappello (changes requested):

icappello (approved):

src/connection.ts
```diff
@@ -235,6 +235,10 @@ export class Connection {
     return this.consumers.size
   }
 
+  public savedFrameMax() {
```
icappello, line 238: maybe let's use a get property

icappello (changes requested):

src/util.ts
```diff
@@ -31,3 +31,6 @@ export function range(count: number): number[] {
   }
   return ret
 }
+
+export const DEFAULT_MAX_FRAME_SIZE = 1048576
+export const DEFAULT_UNLIMITED_FRAME_SIZE = 0
```
albertobarrila, line 36: same here DEFAULT_UNLIMITED_MAX_FRAME or DEFAULT_UNLIMITED_MAX_FRAME_SIZE if the size part is relevant

test/support/fake_data.ts
```diff
@@ -32,14 +32,14 @@ export async function createPublisher(streamName: string, connection: Connection
...
-export function createConnection(username: string, password: string, listeners?: ListenersParams) {
+export function createConnection(username: string, password: string, listeners?: ListenersParams, frameMax?: number) {
   return connect({
     hostname: "localhost",
     port: 5552,
     username,
     password,
     vhost: "/",
-    frameMax: 0, // not used
+    frameMax: frameMax ?? 0, // not used
```
albertobarrila, line 42: remove the comment, now the poor frameMax is used ;-)

albertobarrila (approved):

icappello (approved):

PR #127, opened 2023-12-20
Manage multi node clusters

tarzacodes:


src/client.ts
```diff
@@ -101,17 +106,17 @@ export class Connection {
     this.compressions.set(compression.getType(), compression)
   }
 
-  static connect(params: ConnectionParams, logger?: Logger): Promise<Connection> {
-    return new Connection(logger ?? new NullLogger(), params).start()
+  static async connect(params: ConnectionParams, logger?: Logger): Promise<Client> {
+    return new Client(logger ?? new NullLogger(), { ...params, hostname: params.hostname }).start()
```
icappello, line 110: could we just use `params` as second parameter?

src/client.ts
```diff
@@ -163,9 +168,18 @@ export class Connection {
   public async close(
     params: { closingCode: number; closingReason: string } = { closingCode: 0, closingReason: "" }
   ): Promise<void> {
-    this.logger.info(`Closing connection ...`)
+    this.logger.info(`Closing client...`)
+    if (this.producerCounts()) {
```
icappello, line 172: is it needed to check for the number of producers/consumers before closing them?

src/client.ts
```diff
@@ -63,9 +64,12 @@ export class Connection {
   private heartbeat: Heartbeat
   private consumerId = 0
   private consumers = new Map<number, StreamConsumer>()
+  private producers = new Map<number, StreamProducer>()
+  private producerConnections = new Map<number, Client>()
```
icappello, line 68: Instead of saving the `connection/client` instance on a separate map, we could add a getter method on the producer (the info is already there)

src/client.ts
tarzacodes: It's mostly in order to avoid confusing logs whenever we close the producers/consumers own Clients, so that they don't print "Closing consumers..." as well

src/client.ts
tarzacodes: Yeah, but the producer is something we hand to the user and I didn't want to let them take a hold of the underlying client D:

src/client.ts
```diff
@@ -513,6 +552,17 @@ export class Connection {
   private getBufferSizeParams(): BufferSizeParams {
     return { maxSize: this.frameMax, ...this.bufferSizeSettings }
   }
+
+  private async initNewClient(streamName: string, leader: boolean): Promise<Client> {
+    const [metadata] = await this.queryMetadata({ streams: [streamName] })
+    const chosenNode = leader ? metadata.leader : sample([metadata.leader, ...(metadata.replicas ?? [])])
+    if (!chosenNode) {
+      // need some help with this error message -- what might be better than this? --tarza (20/12/23)
+      throw new Error(`Stream was not found on any node`)
```
icappello, line 561: I think a `Stream not found` message is ok

... so could we remove the comment and the `tarza` entry in the cspell configuration file?

.github/workflows/main.yml
```diff
@@ -21,8 +21,9 @@ jobs:
     services:
       rabbitmq:
         image: rabbitmq:3-management
+        options: --hostname test-node
         env:
-          RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS: -rabbitmq_stream advertised_host localhost
+          # RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS: -rabbitmq_stream advertised_host localhost
```
icappello, line 26: I think we can remove this line

icappello (commented):
just a couple of housekeeping suggestions

src/client.ts
tarzacodes: Done!

.github/workflows/main.yml
tarzacodes: Done!

albertobarrila (approved):
LGTM

PR #135, opened 2024-01-04
Working with a load balancer

tarzacodes:
See https://blog.rabbitmq.com/posts/2021/07/connecting-to-streams/ - part "Client Workaround With a Load Balancer"
Basically, if we are dealing with a load balancer we cannot simply get the node and use the `advertised_host` and `advertised_port`: they might be unreachable for anyone BUT the load balancer. We need to connect to the load balancer and connect to a node through it. Then check if the node we got from the LB is what we wanted (a leader for a Producer, a replica for a Consumer, anything goes for a Locator) and if it is not, close the connection and retry.
**_ADDENDUM: All the clients I've looked thru (Java, Go, dotnet) do the matching with the advertised_host / advertised_port and I've unable to find a way to extract the *actual* hosts and port, therefore at least for now we shall match with those._**
**THEREFORE THIS NO LONGER APPLIES:** 
We'll use the host to match - if a LB is used whoever set up the Rabbit cluster should not have configured `advertised_host` and `advertised_port` and the match should be done on the host name and port.


icappello (dismissed):

l4mby (approved):

icappello (approved):
lgtm

albertobarrila (approved):
lgtm

PR #136, opened 2024-01-04
Basic support for ExchangeCommandVersions messages

### LINKED ISSUE #34: Add exchange command
Ref: https://github.com/rabbitmq/rabbitmq-server/pull/5321

icappello:
closes #34 

Supported commands and related supported versions are sent to the server in the handshake phase, after the `open` command.

The returned values from the server are checked and saved in the `connection` (soon renamed as `client`) instance. If an issue arises (version mismatch between server and client definitions) an error is logged, but no further action is taken.

Note that for the protocol, "no declaration" means "version 1 supported".


src/connection.ts
```diff
@@ -369,6 +374,16 @@ export class Connection {
     })
   }
 
+  private async exchangeCommandVersions() {
+    const versions = clientSupportedVersions
+    const response = await this.sendAndWait<ExchangeCommandVersionsResponse>(
+      new ExchangeCommandVersionsRequest(versions)
+    )
+    response.serverDeclaredVersions.forEach((version) => this.serverDeclaredVersions.push(version))
```
tarzacodes, line 382: Could this be made cleaner with
```this.serverDeclaredVersions.push(...response.serverDeclaredVersions)```
?

File: src/versions.ts [hunk 2]
```diff
@@ -0,0 +1,108 @@
...
+
+  return result
+}
+
+export function checkServerDeclaredVersions(serverDeclaredVersions: Version[], logger: Logger) {
+  const indexedClientVersions = indexVersions(clientSupportedVersions)
+  const indexedServerVersions = indexVersions(serverDeclaredVersions)
+  return (
+    checkVersions(indexedClientVersions, indexedServerVersions, logger) &&
+    checkVersions(indexedServerVersions, indexedClientVersions, logger)
```
tarzacodes, line 106: Just checking: do we need to run checkVersions both ways because one of the two maps might be missing entries? Since we are running the for on the entries of the first Map?

tarzacodes (approved):
LGTM for me

albertobarrila (approved):
LGTM

PR #137, opened 2024-01-08
104 disconnect hook (clean)

### LINKED ISSUE #104: disconnet hook
there's the need to expose a hook for the disconnection of the socket to give the user the ability to manage the situation

icappello:
Closes #104 

On socket disconnect, a user-provided function can be called, if specified.

This callback, linked to the underlying socket event, can be specified:

for locator clients, among the listeners field passed to the Client.connect function. Such a callback is specific for the locator connection and is not linked to connection events related to publisher/consumer instances created from the locator.
for publishers and consumers, as a field in the parameters of declarePublisher and declareConsumer. Such callbacks are specific for the single publisher/consumer.


tarzacodes (approved):
LGTM

PR #141, opened 2024-01-09
Rename `sendChunk` to `sendBuffer`

### LINKED ISSUE #140: Rename function `sendChunk` in `Producer` class
To avoid confusion, `sendChunk` should be renamed to something like `sendBuffer`, since "chunk" has another meaning in the rabbitmq context

icappello:
closes #140


tarzacodes (approved):
LGTM

PR #148, opened 2024-01-12
20-feature-single-active-consumer

### LINKED ISSUE #20: Feature: Single Active Consumer
Implement the Single-Active Consumer feature.
Java reference: https://rabbitmq.github.io/rabbitmq-stream-java-client/sac/htmlsingle/#single-active-consumer

Starting from: RabbitMQ 3.11.0

l4mby:
Adds possibility of declaring a single active consumer.
Closes #20


src/consumer.ts
```diff
@@ -25,12 +27,14 @@ export class StreamConsumer implements Consumer {
...
       consumerId: number
       consumerRef?: string
+      offset: Offset
     }
   ) {
     this.client = params.client
     this.stream = params.stream
     this.consumerId = params.consumerId
     this.consumerRef = params.consumerRef
+    this.offset = params.offset
```
tarzacodes, line 37: I don't get why you added the offset property inside the Consumer, where is it used?

src/responses/responses.ts
```diff
@@ -22,3 +22,5 @@ export { StreamStatsResponse } from "./stream_stats_response"
 export { SubscribeResponse } from "./subscribe_response"
 export { TuneResponse } from "./tune_response"
 export { UnsubscribeResponse } from "./unsubscribe_response"
+export { ConsumerUpdateQuery } from "./consumer_update_query"
+export { ConsumerUpdateResponse } from "../requests/consumer_update_response"
```
tarzacodes, line 26: Is this mixing necessary? I'd just export the response one from the requests

src/consumer.ts
l4mby: we need the offset when sending the ConsumerUpdateResponse, so that we complete the promotion of single active consumers

test/e2e/declare_consumer.test.ts
```diff
@@ -59,6 +59,90 @@ describe("declare consumer", () => {
     await eventually(() => expect(messages).eql([Buffer.from("hello")]))
   }).timeout(10000)
 
+  it("declaring multiple active consumer on an existing stream - only one consumer should handle the message", async () => {
```
tarzacodes, line 62: multiple -single- active consumers

src/consumer.ts
tarzacodes: Right, my bad!

tarzacodes (approved):
LGTM

PR #149, opened 2024-01-12
59 ability to consume messages from superstream

tarzacodes:


icappello (approved):
lgtm

src/client.ts
```diff
@@ -310,6 +321,11 @@ export class Client {
     return res.ok
   }
 
+  public async declareSuperStreamConsumer(superStream: string, handle: ConsumerFunc): Promise<SuperStreamConsumer> {
+    const partitions = await this.queryPartitions({ superStream })
+    return SuperStreamConsumer.create(handle, { locator: this, consumerRef: "test", partitions })
```
albertobarrila, line 326: the consumerRef is hardcoded as "test", is it right?

src/publisher.ts
```diff
@@ -82,6 +82,7 @@ export class StreamPublisher implements Publisher {
   private scheduled: NodeJS.Immediate | null
   private logger: Logger
   private maxChunkLength: number
+  private isClosed = false
```
albertobarrila, line 85: I honestly do not like "is" properties, I would suggest using only closed, wdyt?

src/super_stream_consumer.ts
```diff
@@ -0,0 +1,53 @@
+import { Client } from "./client"
+import { Consumer, ConsumerFunc } from "./consumer"
+import { Offset } from "./requests/subscribe_request"
+
+export class SuperStreamConsumer {
+  private consumers: Map<string, Consumer> = new Map<string, Consumer>()
+  public consumerRef: string
+  private locator: Client
```
albertobarrila, line 8: is locator a good name for the client underneath the consumer?

tarzacodes: It is the locator, that's the term we are using for the client that manages the connections right?

PR #152, opened 2024-01-12
142-docs-example-for-cluster-and-load-balancer

l4mby:
Add new information on the readme


README.md
```diff
@@ -104,17 +169,81 @@ const client = await connect({
...
+```typescript
+const firstNode = "node0:5562" // can be taken from envs
+const resolver = getAddressResolverFromEnv()
+const client = await connect({
+  hostname: "node0",
+  port: 5562,
+  username: "rabbit",
+  password: "rabbit",
+  vhost: "/",
+  addressResolver: { enabled: true, endpoint: resolver },
```
albertobarrila, line 226: here we have addressResolver whit enabled and endpoint @tarzacodes can we make more ergonomic using hostname and port from the outside and collapsing it in just a boolean?

tarzacodes: Are you sure? I think that it'd be quite limiting if we were to force people to put their resolver in the env vars.... maybe they want to relate it to some other endpoint or get it from the network or any other number of options. Might be reasonable, if enabled is true and no endpoint is specified, to read the env vars as a default, WDYT?

tarzacodes: Ok, I made it so that if the endpoint is not defined it'll use the client hostname/port

tarzacodes (approved):
LGTM

PR #154, opened 2024-01-15
[WIP] Reuse underlying clients/sockets when possible

icappello:
If the client application creates multiple producers for the same stream or if it creates multiple consumers for the same stream and these happen to insist on the same node, then it makes sense to reuse the same underlying Client instance (and by extension the same tcp socket).

This behaviour should anyway be limited (e.g. at most 10 producers share the same Client) and it must not change the behavior of the communication protocol


src/client.ts
```diff
@@ -94,6 +110,9 @@ export class Client {
       this.socket.connect(this.params.port, this.params.hostname)
     }
     if (params.socketTimeout) this.socket.setTimeout(params.socketTimeout)
+    this.hostname = params.hostname
+    this.leader = params.leader || false
```
tarzacodes, line 114: Useful tip: to define defaults, use ?? instead of ||. In this specific instance nothing changes, but if the default were `true` and params.leader was in fact false, `params.leader || true` evaluates as true while `params.leader ?? true` evaluates as false. Some linter settings even yell at you if you use || this way

src/util.ts
```diff
@@ -15,6 +15,7 @@ export function range(count: number): number[] {
 
 export const DEFAULT_FRAME_MAX = 1048576
 export const DEFAULT_UNLIMITED_FRAME_MAX = 0
+export const MAX_SHARED_CLIENT_INSTANCES = +(process.env.MAX_SHARED_CLIENT_INSTANCES || 10)
```
tarzacodes, line 19: Better use ?? again, and 10 might be a little low. I think that the dotnet uses somewhere around 200, maybe we could do with 100 at least

src/client.ts
```diff
@@ -695,6 +732,49 @@ export class Client {
...
+
+  private incrRefCount() {
+    return ++this.refCount
+  }
+
+  private decrRefCount() {
+    return --this.refCount
+  }
+
+  private static GetUsableCachedClient(leader: boolean, streamName: string, host: string) {
```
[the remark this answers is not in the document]
icappello: old habits die hard...

tarzacodes (approved):
LGTM for me

tarzacodes (approved):
LGTM for me

src/client.ts
```diff
@@ -772,6 +816,48 @@ export class Client {
     bindingKeys.map((bk) => partitions.push(`${streamName}-${bk}`))
     return { partitions, streamBindingKeys: bindingKeys }
   }
+  private incrRefCount() {
+    return ++this.refCount
```
albertobarrila, line 820: the return value is used? otherwise i'll remove the return

src/client.ts
```diff
@@ -772,6 +816,48 @@ export class Client {
...
+  }
+
+  private decrRefCount() {
+    return --this.refCount
```
albertobarrila, line 824: same here

src/client.ts
```diff
@@ -194,21 +220,25 @@ export class Client {
...
+      if (this.consumerCounts()) {
+        this.logger.info(`Stopping all consumers...`)
+        await this.closeAllConsumers()
+      }
+      this.logger.info(`Stopping heartbeat...`)
+      this.heartbeat.stop()
+      this.logger.debug(`Close...`)
+      const closeResponse = await this.sendAndWait<CloseResponse>(new CloseRequest(params))
+      this.logger.debug(`Close response: ${closeResponse.ok} - '${inspect(closeResponse)}'`)
+      Client.removeCachedClient(this)
```
albertobarrila, line 239: having a static method that take the instance as a parameter seems odd but I think that is a smell of client having too much responsibilities

src/client.ts
icappello: ok, I'll remove the return statement

src/client.ts
icappello: this one is used when `close` is invoked to decide if the cleanup process must take place or not

albertobarrila (approved):
lgtm

PR #155, opened 2024-01-16
105 - Create/Delete Super Stream

l4mby:
Requires rabbitmq:3.13-rc-management, so that the server can parse the commands


package.json
```diff
@@ -53,5 +53,8 @@
     "ts-node": "^10.9.1",
     "typescript": "^4.9.5",
     "winston": "^3.8.2"
+  },
+  "dependencies": {
+    "semver": "^7.5.4"
```
tarzacodes, line 58: Adding a dep is quite a commitment, are we 100% we need this?

l4mby: We use it to compare the rabbitmq management versions, maybe we can try to hardcode the comparison?

tarzacodes (approved):
LGTM for me, @icappello ?

icappello (approved):
looks good to me

PR #157, opened 2024-01-16
153 59 ability to consume messages from superstream typos

tarzacodes:
I did not rename "locator" since I could not find a more fitting name. `isClosed` was renamed to `closed` and the `consumerRef` for the SuperStream consumer is now randomly assigned


albertobarrila (approved):
LGTM

PR #165, opened 2024-01-23
Split client responsibilities

icappello:
Class ConnectionProxy represents now an intermediate level between the Client and the socket

Class ConnectionProxyPool handles the ConnectionProxy instances for resource sharing


albertobarrila (commented):
waht about renaming connection_proxy to connection and connection_proxy_pool to connection_pool? for the rest lgtm

albertobarrila (dismissed):
lgtm

src/connection_pool.ts
```diff
@@ -0,0 +1,52 @@
...
+      ConnectionPool.removeCachedConnectionProxy(connectionProxy)
+      return true
+    }
+    return false
+  }
+
+  public static removeCachedConnectionProxy(connectionProxy: Connection) {
+    const leader = connectionProxy.leader
+    const streamName = connectionProxy.streamName
+    const host = connectionProxy.hostname
```
tarzacodes, line 38: Can be rewritten as 
```
const { leader, streamName, hostname: host } = connectionProxy
```

src/connection.ts
```diff
@@ -0,0 +1,457 @@
...
+function extractHeartbeatInterval(heartbeatInterval: number, tuneResponse: TuneResponse): number {
+  return heartbeatInterval === 0 ? tuneResponse.heartbeat : Math.min(heartbeatInterval, tuneResponse.heartbeat)
+}
+
+export class Connection {
+  public readonly hostname: string
+  public readonly leader: boolean
+  public readonly streamName: string | undefined
+  private socket: Socket
+  private correlationId = 100
```
tarzacodes, line 71: I know it's not relevant to the PR (it was already there) but do we know why we have this default?

icappello: I'm not sure. @albertobarrila do you have some insight on this? Anyway, I think it would be better to address this in another issue, if we decide to change it

tarzacodes (approved):
LGTM for me as well

PR #168, opened 2024-01-24
feat: publishing in batch and compression on superstream

tarzacodes:
Actually contains also a test for testing the reading of compressed messages on superstream, closing issues 56 and 58 both

### DIFF
--- a/src/super_stream_publisher.ts
+++ b/src/super_stream_publisher.ts
@@ -1,6 +1,7 @@
 import { Client, RoutingStrategy } from "./client"
+import { CompressionType } from "./compression"
 import { murmur32 } from "./hash/murmur32"
-import { MessageOptions, Publisher } from "./publisher"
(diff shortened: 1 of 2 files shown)


src/super_stream_publisher.ts
```diff
@@ -57,6 +59,37 @@ export class SuperStreamPublisher {
...
+        if (!msgs) {
+          msgs = []
+          messagesByPartition.set(partition, msgs)
+        }
+        msgs.push(m)
+      })
+    )
+
+    // init all publishers
+    const publishersMap: Map<string, Publisher> = new Map()
```
icappello, line 78: Is `publishersMap` needed?

tarzacodes: Actually, not. Good spot, thanks

icappello (approved):

PR #171, opened 2024-01-25
Adds filtering feature on publishing side

l4mby:


src/client.ts
```diff
@@ -427,6 +440,7 @@ export class Client {
 
   public async createStream(params: { stream: string; arguments?: CreateStreamArguments }): Promise<true> {
     this.logger.debug(`Create Stream...`)
+    this.filteringEnabled = lt(coerce(this.rabbitManagementVersion)!, REQUIRED_MANAGEMENT_VERSION) ? false : true
```
icappello, line 443: I suggest to move this assignment in the connect callback, so it is performed as soon as the property is available

src/publisher.ts
```diff
@@ -182,6 +187,12 @@ export class StreamPublisher implements Publisher {
   }
 
   private async enqueue(publishRequestMessage: PublishRequestMessage) {
+    if (this.filter) {
+      publishRequestMessage.filterValue = this.filter(publishRequestMessage.message)
+    }
+    if (!this.client.isFilteringEnabled && this.filter) {
```
icappello, line 193: I think the error should be thrown only if `this.filter` is defined (let's use the short-circuit evaluation)

```suggestion
    if (this.filter && !this.client.isFilteringEnabled) {
```

src/response_decoder.ts
```diff
@@ -115,7 +120,7 @@ function decodeResponse(
 ): PossibleRawResponses {
   const key = dataResponse.readUInt16()
   const version = dataResponse.readUInt16()
-  if (key === DeliverResponse.key) {
+  if (key === DeliverResponse.key && version === 1) {
```
tarzacodes, line 123: Maybe we could go with a single if branch? Most of the content is the same

src/client.ts
```diff
@@ -137,16 +138,23 @@ export class Client {
...
-      connection: connection,
-      stream: params.stream,
-      publisherId: publisherId,
-      publisherRef: params.publisherRef,
-      boot: params.boot,
-      maxFrameSize: this.maxFrameSize,
-      maxChunkLength: params.maxChunkLength,
-      logger: this.logger,
-    })
+    if (filter && !connection.isFilteringEnabled) {
```
tarzacodes, line 141: Isn't closing the connection a bit too harsh?

tarzacodes (approved):
LGTM

tarzacodes (approved):
LGTM

PR #172, opened 2024-01-25
Adds filtering feature on consumer side

l4mby:
Do not close before #171


src/client.ts
```diff
@@ -194,6 +203,12 @@ export class Client {
       properties["single-active-consumer"] = "true"
       properties["name"] = params.consumerRef!
     }
+    if (params.filter) {
+      for (let i = 0; i < params.filter.values.length; i++) {
+        properties[`${i}`] = params.filter.values[i]
```
icappello, line 208: Is the key `${i}` correct? In the docs it is specified as `filter.0`, `filter.1` and as far as I understand, these properties should set a filter on the broker side (I may be wrong)

icappello (approved):

PR #176, opened 2024-01-30
[bug] creation of stream with arguments

l4mby:
Creation stream arguments were ignored because the property key started with `x-`.
Moreover the  `queue-leader-locator` argument can take up to 3 values which are `random`, `client-local` and `least-leaders`.
Also, the `max-age` argument is a string of numbers ending with a trailing `s`

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ await client.close()
 
 ### Filtering
 
-It is possible to filter messages both on the publishing and consumer sides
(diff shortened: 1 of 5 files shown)


icappello (dismissed):
lgtm

tarzacodes (dismissed):
LGTM for me as well!

tarzacodes (approved):
LGTM again

PR #189, opened 2024-02-13
Update readme and add example to show how one might manage auto reconnecting

### LINKED ISSUE #187: Update README and add example to show how one might manage auto-reconnecting
We should add an example that shows one correct way to handle auto reconnecting whenever a connection is closed by the server or due to a Metadata Update.
That example should also be referenced in the README

icappello:
closes #187


tarzacodes (approved):
Brilliant, couldn't find anything wrong with it! LGTM for me!

Issue #190, opened 2024-03-27
Allow Customized Consumer Names and Offset for Super Stream Consumers


haile-vnm:
## Current Behavior
- Currently, when consuming a super stream, the following limitations exist:
  - The Single Active Consumer feature cannot be enabled due to the random generation of consumer names.
  - The consuming point in the super stream is fixed to start from the beginning (`Offset.first()`).

## Expected Behavior
- The expected behavior includes the following improvements:
  - Allow the Single Active Consumer feature by providing a customized consumer reference (`consumerRef`) when declaring the super stream consumer. This would enable users to specify their own consumer names instead of relying on random generation.
  - Allow users to decide the consuming point by providing the option to set the offset when declaring the Single Active Consumer. This would give users the ability to choose a specific point in the super stream from which consumption should begin.

icappello:
Hello, thanks for the opened issue and the PR, much appreciated!

We will give you a feedback shortly.

icappello:
Hello @haile-vnm your contribution has been merged into the codebase, so I'll now close this issue. Thanks again for opening the issue and providing an implementation!

PR #191, opened 2024-03-27
feat(super-stream): enable customized consumerRef, offset

haile-vnm:
### ⛔Before
```js
client.declareSuperStreamConsumer(
      {
        superStream: superStreamName
        // consumerRef: "super-stream-[random-uuid]", as default, can not change
        // offset: Offset.first(), as default, can not change
      },
      (message: Message) => {
        // handle message logic
      })
```
### ✅After
```js
client.declareSuperStreamConsumer(
      {
        superStream: superStreamName,
        consumerRef: "consumer-group-name", // defined by user
        offset: Offset.next() // defined by user
      },
      (message: Message) => {
        // handle message logic
      })
```

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -207,12 +207,16 @@ export class Client {
   }
 
   public async declareSuperStreamConsumer(
-    { superStream }: DeclareSuperStreamConsumerParams,
+    { superStream, offset, consumerRef }: DeclareSuperStreamConsumerParams,
     handle: ConsumerFunc
   ): Promise<SuperStreamConsumer> {
-    const consumerRef = `${superStream}-${randomUUID()}`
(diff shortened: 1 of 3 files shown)


haile-vnm:
https://github.com/coders51/rabbitmq-stream-js-client/issues/190

icappello:
Hello, sorry for the delay.
We were unsure at first about allowing any type of offset for super-streams, in particular `numeric` and `last` offsets since the result on the consumer side depends on the number of partitions. That said, after a bit of internal discussion we agree on this behavior.

One note: the CI pipeline fails when performing the format check on a test file. Could you run a `npm run format` and update the PR?

haile-vnm:
Hi @icappello , 

I fixed the format. Please help me to check it again. Thanks!

PR #194, opened 2024-05-30
Check if cached connection can support more pub/subs before reusing

moughxyz:
Hello,

A single connection can only support 255 publishers or consumers, given that the publisherId is an 8 bit integer.

https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_stream/docs/PROTOCOL.adoc#declarepublisher

Other libraries like Go and Java create a new connection when this limit is reached: https://groups.google.com/g/rabbitmq-users/c/MQ-Bub3q13c/m/9-3cKfpbAAAJ

The issue with the current implementation is that the cached connection is always used when creating publishers, and this means that on the 256th publisher, it will crash because the publisherId of 256 won't fit into the 8 bit slot.

In this PR, we check to see if a connection can first support an additional publisher. If not, we create a new connection.

This would be a first-stab implementation. Please let me know any shortcomings and would be happy to fix.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -54,6 +54,9 @@ type PublisherMappedValue = {
   params: DeclarePublisherParams
   filter: FilterFunc | undefined
 }
+
+const MAX_PUBLISHERS_OR_SUBSCRIBERS = 255
+
 export class Client {
   public readonly id: string = randomUUID()
   private publisherId = 0
@@ -62,6 +65,8 @@ export class Client {
   private publishers = new Map<number, PublisherMappedValue>()
   private compressions = new Map<CompressionType, Compression>()
   private connection: Connection
+  private currentPublisherConnectionIndex = 0
+  private currentConsumerConnectionIndex = 0
 
   private constructor(private readonly logger: Logger, private readonly params: ClientParams) {
     this.compressions.set(CompressionType.None, NoneCompression.create())
@@ -150,7 +155,7 @@ export class Client {
     const publisher = new StreamPublisher(streamPublisherParams, filter)
     this.publishers.set(publisherId, { publisher, connection, params, filter })
     this.logger.info(
-      `New publisher created with stream name ${params.stream}, publisher id ${publisherId} and publisher reference ${params.publisherRef}`
+      `New publisher created with stream name ${params.stream}, publisher id ${publisherId}, connection index ${this.currentPublisherConnectionIndex} and publisher reference ${params.publisherRef}`
     )
     return publisher
   }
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -182,10 +187,12 @@ export class Client {
...
-    await this.declareConsumerOnConnection(params, consumerId, this.connection)
-    this.logger.info(
-      `New consumer created with stream name ${params.stream}, consumer id ${consumerId} and offset ${params.offset.type}`
-    )
+    try {
+      await this.declareConsumerOnConnection(params, consumerId, this.connection)
+      this.logger.info(
+        `New consumer created with stream name ${params.stream}, consumer id ${consumerId}, connection index ${this.currentConsumerConnectionIndex} and offset ${params.offset.type}`
+      )
+    } catch (error) {}
```
tarzacodes, line 195: Is it wise to catch the error here and do nothing? What happens if we fail to declare the consumer on the connection?

src/client.ts
```diff
@@ -467,13 +475,19 @@ export class Client {
 
   private incPublisherId() {
     const publisherId = this.publisherId
-    this.publisherId++
+    if (++this.publisherId >= MAX_PUBLISHERS_OR_SUBSCRIBERS) {
```
tarzacodes, line 478: I'm not sure this would be the best approach, since this is an always increasing index and the publishers (as the consumers) can also be destroyed. We keep, on the client, a map of active consumers/publishers for other purposes: we could use the size of the map in order to determine whether we open a new connection or not. Or maybe you already considered that and found some issue with this approach that I'm failing to spot right now, let me know,

src/client.ts
```diff
@@ -582,17 +607,18 @@ export class Client {
...
   private async getConnectionOnChosenNode(
     leader: boolean,
+    index: number,
     streamName: string,
     chosenNode: { host: string; port: number },
     metadata: StreamMetadata,
     connectionClosedListener?: ConnectionClosedListener
   ): Promise<Connection> {
-    const connectionParams = this.buildConnectionParams(leader, streamName, connectionClosedListener)
+    const connectionParams = this.buildConnectionParams(leader, streamName, index, connectionClosedListener)
```
tarzacodes, line 621: Maybe we don't need to store the index inside of the connection? It looks like we are just using this index in the caching part, which the connection itself should not be aware of... Could we simply pass it as an arg to cacheConnection?

moughxyz:
Hey @tarzacodes, apologies, I have since reevaluated our usage of RabbitMQ and may go in a different direction. So I'm afraid I can no longer give this PR the care it deserves. I will close it in the meantime and perhaps it can serve as a reference to anyone else encountering this issue.

PR #196, opened 2024-06-12
Add configurable policies for consumer credit flow

icappello:


src/consumer_credit_policy.ts
```diff
@@ -0,0 +1,65 @@
...
+    await requestWrapper(this.onRenewal)
+  }
+
+  public onSubscription(): number {
+    return this.onStartup
+  }
+}
+
+class NewCreditsOnChunkProgress extends ConsumerCreditPolicy {
+  constructor(onStartup: number = 1, private readonly ratio: number = 0.5, private readonly howMany: number = 1) {
```
tarzacodes, line 36: Maybe rename `howMany` to `onRenewal`? It would be more consistent and it's easier to understand what is happening, at least for me.

src/consumer_credit_policy.ts
```diff
@@ -0,0 +1,65 @@
...
+    const threshold = Math.max(1, Math.ceil(this.ratio * total))
+
+    if (current === threshold) {
+      await requestWrapper(this.howMany)
+    }
+  }
+}
+
+class NewCreditsOnChunkCompleted extends ConsumerCreditPolicy {
+  constructor(onStartup: number = 1, private readonly howMany: number = 1) {
```
tarzacodes, line 50: Same, howMany -> onRenewal would be a little be easier to read, IMHO

src/consumer.ts
```diff
@@ -22,6 +23,7 @@ export class StreamConsumer implements Consumer {
   public consumerRef?: string
   public offset: Offset
   private clientLocalOffset: Offset
+  private creditsHandler: ConsumerCreditPolicy
```
tarzacodes, line 26: If its named policy everywhere else, might be apt to rename the class property

src/client.ts
```diff
@@ -499,8 +516,14 @@ export class Client {
       }
       this.logger.debug(`on deliverV1 -> ${consumer.consumerRef}`)
       this.logger.debug(`response.messages.length: ${response.messages.length}`)
-      await this.askForCredit({ credit: 1, subscriptionId: response.subscriptionId }, connection)
-      response.messages.map((x) => consumer.handle(x))
+      const creditRequestWrapper = this.askForCredit(response.subscriptionId, connection)
```
tarzacodes, line 519: Since this part is identical for v1 and v2 (maybe I'm wrong), should we extract it to a separate method taking the response, connection and consumer?

test/e2e/consumer_credit_flow_policy.test.ts
```diff
@@ -0,0 +1,170 @@
...
+    const consumerRef = createConsumerRef()
+    const policy = creditsOnChunkReceived(1, 1)
+    sandbox.on(policy, "onChunkReceived")
+    const received: Message[] = []
+    await client.declareConsumer(
+      {
+        stream: streamName,
+        offset: Offset.first(),
+        singleActive: true,
+        consumerRef: consumerRef,
```
tarzacodes, line 81: `consumerRef: consumerRef,` can be substituted with `consumerRef,` due to TS provided syntactic sugar

test/e2e/consumer_credit_flow_policy.test.ts
```diff
@@ -0,0 +1,170 @@
...
+    )
+
+    await send(publisher, chunks)
+
+    await eventually(() => expect(received.length).eql(nMessages))
+    await eventually(() => expect(policy.onChunkReceived).called.exactly(chunks.length))
+    await always(() => expect(policy.onChunkReceived).called.below(chunks.length + 1), 5000)
+  }).timeout(10000)
+
+  it("NewCreditOnChunkProgress policy requests new credit when chunk handled at 50% progress", async () => {
```
tarzacodes, line 96: Since we use tests as documentation, it might be be better to be even more explicit - something like `NewCreditOnChunkProgress policy requests new credit when chunk handled at 50% progress when the set ratio is 0.5`.

test/e2e/consumer_credit_flow_policy.test.ts
```diff
@@ -0,0 +1,170 @@
...
+    )
+
+    await send(publisher, chunks)
+
+    await eventually(() => expect(received.length).eql(nMessages))
+    await eventually(() => expect(policy.onChunkProgress).called.exactly(nMessages))
+    await always(() => expect(policy.onChunkProgress).called.below(nMessages + 1), 5000)
+  }).timeout(10000)
+
+  it("NewCreditOnChunkProgress policy requests new credit when chunk handled at 100% progress", async () => {
```
tarzacodes, line 121: Maybe we can measure the timing by having the consumer callback set a flag if the `received` becomes bigger than the ratio*length before the spy has been called? Basically, measure that, if ratio is 0.5, we don't have 0.6*length messages handled before the spy has been called, and if so set a flag we can expect to.

test/e2e/consumer_credit_flow_policy.test.ts
```diff
@@ -0,0 +1,170 @@
...
+    )
+
+    await send(publisher, chunks)
+
+    await eventually(() => expect(received.length).eql(nMessages))
+    await eventually(() => expect(policy.onChunkProgress).called.exactly(nMessages))
+    await always(() => expect(policy.onChunkProgress).called.below(nMessages + 1), 5000)
+  }).timeout(10000)
+
+  it("NewCreditOnChunkCompleted policy requests new credit when chunk completely handled", async () => {
```
tarzacodes, line 146: Same as before, we can measure that the spy has been called only AFTER all the messages have been handled. If we don't do this, we have a set of actually identical tests which might be problematic.

test/e2e/declare_consumer.test.ts
```diff
@@ -178,9 +179,12 @@ describe("declare consumer", () => {
 
   it(`consume a lot of messages`, async () => {
     const receivedMessages: Buffer[] = []
-    await client.declareConsumer({ stream: streamName, offset: Offset.next() }, (message: Message) => {
-      receivedMessages.push(message.content)
-    })
+    await client.declareConsumer(
+      { stream: streamName, offset: Offset.next(), creditPolicy: creditsOnChunkReceived(1, 1) },
```
tarzacodes, line 183: Why do we have to set a non-default `creditPolicy` here?

test/e2e/shared_publisher_clients.test.ts
```diff
@@ -64,9 +65,12 @@ describe("publish messages through multiple publishers", () => {
       }
     }
 
-    await client.declareConsumer({ stream: streamName, offset: Offset.first() }, (msg) => {
-      received.push(msg.content.toString("utf-8"))
-    })
+    await client.declareConsumer(
+      { stream: streamName, offset: Offset.first(), creditPolicy: creditsOnChunkReceived(1, 1) },
```
tarzacodes, line 69: Same as other comment, is the non-default creditPolicy important here? Should that be documented? Or is it only for performance reasons?

src/client.ts
icappello: We can refactor now or follow the rule of three, it's the same to me

tarzacodes: Touche'

test/e2e/declare_consumer.test.ts
icappello: TLDR: 
The default values for credits are now set to have two credits on startup and one credit on renewal.
Given how the node event loop handles async calls, and how the test is written, either we need to use exacly one credit for this test, we change the checked condition, or we change the default values.

Long version:

First of all, note that in the checked condition for this test, the order in which messages are handled by the callback matters.

In the previous implementation (https://github.com/coders51/rabbitmq-stream-js-client/blob/16d26ff6ccc03485e00388fbdab3c506caea077d/src/client.ts) on startup 10 credits are assigned to the client. The message handling is implemented using a `.map` call, which does not allow for async behavior; for the node event loop this is basically an atomic synchronous call.

In this PR this has been updated with a `for{}` cycle to allow the async call to `onChunkProgress`.

The presence of this async call is enough for the event loop to allow the handling of received chunks to be interleaved. Note that only the handling on the client side is affected, the delivery order from RabbitMQ remains consistent with what was received from the publisher.

Let's start from the code in the main branch (so the new added logic is not involved) and substitute the call
```
response.messages.map((x) => consumer.handle(x))
```

to something like

```
for (const m of response.messages) {
    consumer.handle(m)
    await new Promise((res) => {
        res(true)
    })
}
```

which at first glance should behave exactly the same; the test will actually fail.

src/client.ts
```diff
@@ -499,8 +516,14 @@ export class Client {
...
+      await consumer.creditPolicy.onChunkReceived(creditRequestWrapper)
+      const chunkLength = response.messages.length
+      for (const [idx, message] of response.messages.entries()) {
+        consumer.handle(message)
+        await consumer.creditPolicy.onChunkProgress(idx + 1, chunkLength, creditRequestWrapper)
+      }
```
gpad, line 525: This code is duplicated with the above and contains a `for` ....

src/client.ts
```diff
@@ -516,12 +539,20 @@ export class Client {
...
       if (consumer.filter) {
         response.messages.filter((x) => consumer.filter?.postFilterFunc(x)).map((x) => consumer.handle(x))
         return
       }
-      response.messages.map((x) => consumer.handle(x))
+
+      for (const [idx, message] of response.messages.entries()) {
+        consumer.handle(message)
+        await consumer.creditPolicy.onChunkProgress(idx + 1, chunkLength, creditRequestWrapper)
+      }
```
gpad, line 554: It's quite similar to above code

src/consumer_credit_policy.ts
gpad: onRenewal for me is a function/handler, (same as startup) usally in node onSoemthig indicate an handler, it's more `startFrom` and `step`

src/consumer_credit_policy.ts
gpad: as above

tarzacodes (approved):

Issue #197, opened 2024-06-13
Check if cached connection can support more pub/subs before reusing


tarzacodes:
Taking from the proposed PR (now closed by the original author) 
https://github.com/coders51/rabbitmq-stream-js-client/pull/194
Implement conditional reusing of cached connections

tarzacodes:
Actually, after reviewing our current implementation I found out that we already supported a hard cap on the number of consumers/publishers on a single connection (and the opening of new connections to host those extra publishers/consumers), which can be set thru an env variable. This is good, but we still had two issues:
1) The publisher/consumer ids are unique across connections (they are kept track of in the client), and they increase monothically. So a client opening 257 publishers, even on different connections, would crash since the publisherId must be <256 (it is a uint8)
2) The max number of consumers/publishers can be set higher than 256
This PR linked to this issue takes care of these two points. Now publisher and consumer ids will be kept track of on the connection object.

PR #198, opened 2024-06-13
197 check if cached connection can support more pubsubs before reusing

tarzacodes:
See Issue https://github.com/coders51/rabbitmq-stream-js-client/issues/197 for more details


src/client.ts
```diff
@@ -549,26 +548,26 @@ export class Client {
 
   private async getConnection(
     streamName: string,
-    leader: boolean,
+    purpose: "publisher" | "consumer",
```
icappello, line 551: I think it would make sense to define a union type and a couple of constants to reduce the use of plain strings

icappello (changes requested):

icappello (approved):

PR #199, opened 2024-06-25
Update to node 20

tarzacodes:


tarzacodes:
Most of the changes are due to linting - the linter got updated

.tool-versions
```diff
@@ -1,2 +1,2 @@
-nodejs 16.15.1
+nodejs 20.11.0
```
albertobarrila, line 1: the last LTS is 20.15.0

albertobarrila (approved):

.github/workflows/main.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16.x]
+        node-version: [20.x]
```
gpad, line 18: Why don't keep a matrix like, [16.x, 18.x, 20.x] ?

example/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  // "extends": "@tsconfig/node16/tsconfig.json",
+  // "extends": "@tsconfig/node-lts/tsconfig.json",
```
gpad, line 2: Why is this commented?

package.json
```diff
@@ -8,7 +8,7 @@
     "build": "tsc",
     "check": "npm run check-ts && npm run check-lint && npm run check-format && npm run check-spell",
     "check-ts": "tsc --noEmit && tsc --project test/tsconfig.json --noEmit",
-    "check-lint": "tsc --noEmit && eslint 'src/**/*.ts' 'test/**/*.ts'",
+    "check-lint": "tsc --noEmit && eslint --fix 'src/**/*.ts' 'test/**/*.ts'",
```
gpad, line 11: The fix change the file it's a check so the --fix should not be added

package.json
```diff
@@ -28,30 +28,30 @@
   },
   "homepage": "https://github.com/coders51/rabbitmq-stream-js-client#readme",
   "devDependencies": {
-    "@tsconfig/node16": "^1.0.3",
+    "@tsconfig/node-lts": "^20.1.1",
```
gpad, line 31: Isn't it deprecated?

performance_test/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  // "extends": "@tsconfig/node16/tsconfig.json",
+  // "extends": "@tsconfig/node-lts/tsconfig.json",
```
gpad, line 2: As above

src/amqp10/encoder.ts
```diff
@@ -47,7 +47,7 @@ type MessageAnnotationsList = { key: string; value: MessageAnnotationsValue }[]
 
 export function amqpEncode(
   writer: DataWriter,
-  { content, messageProperties, applicationProperties, messageAnnotations }: Message
+  { content, messageProperties, applicationProperties, messageAnnotations }: Message,
```
gpad, line 50: If you set a property in the prettierc 
```
  "trailingComma": "es5"
```
You don't have to add this orrible comma and you don't have to change all these files

example/tsconfig.json
tarzacodes: Dunno, found it like this - I'll clean it up since it doesn't seem to be of use

package.json
tarzacodes: ...is it? I couldn't find anything on google, do you happen to have some source I can look into?

albertobarrila (approved):

PR #207, opened 2024-07-24
feat: expose query offset in the client

l4mby:


icappello (dismissed):

test/e2e/offset.test.ts
```diff
@@ -204,23 +204,34 @@ describe("offset", () => {
...
+        { stream: testStreamName, offset: Offset.first(), consumerRef: "my_consumer" },
         async (message: Message) => {
-          await consumer.storeOffset(message.offset!)
-          offset = message.offset!
+          if (message.content.toString() === "marker") {
+            await consumer.storeOffset(message.offset!)
+          }
+        }
+      )
+      await wait(3000)
```
gpad, line 223: Instead of a fix wait that is always to short or too long, I suggest to find a trigger or for example set a flag that trigger the other behaviour

tarzacodes (approved):
LGTM

PR #213, opened 2024-08-28
Add support for EXTERNAL auth

david-mohr:
Adds an optional `mechanism` parameter and implements EXTERNAL auth following the Java reference implementation.

Also fixes a broken test.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -707,7 +707,7 @@ export interface ClientParams {
   port: number
   username: string
   password: string
-  mechanism?: string
+  mechanism?: "PLAIN" | "EXTERNAL"
   vhost: string
   frameMax?: number
   heartbeat?: number
--- a/src/client.ts
+++ b/src/client.ts
@@ -707,6 +707,7 @@ export interface ClientParams {
   port: number
   username: string
   password: string
+  mechanism?: string
   vhost: string
   frameMax?: number
   heartbeat?: number
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -146,7 +146,11 @@ export class Connection {
         this.logger.info(`Connected to RabbitMQ ${this.params.hostname}:${this.params.port}`)
         this.peerProperties = (await this.exchangeProperties()).properties
         this.filteringEnabled = lt(coerce(this.rabbitManagementVersion)!, REQUIRED_MANAGEMENT_VERSION) ? false : true
-        await this.auth({ username: this.params.username, password: this.params.password })
+        await this.auth({
+          username: this.params.username,
+          password: this.params.password,
+          mechanism: this.params.mechanism ?? "PLAIN",
+        })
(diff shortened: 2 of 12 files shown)


icappello:
Hello, thanks for opening this PR! We will give you a feedback in the next few days

src/connection.ts
icappello, line 1: Could you add a couple of unit/e2e tests for this feature? What changes or updates would be needed to the docker/docker-compose definitions?

test/support/util.ts
icappello, line 1: in case it is needed, please use the ENV variable to override the static configuration

david-mohr: the static configuration was causing the existing tests to fail, which is why I changed it. I can remove this change.

src/connection.ts
david-mohr: Rabbit offers a method to generate some tests keys: https://www.rabbitmq.com/docs/ssl#automated-certificate-generation. It can probably be used for testing, I will investigate

david-mohr:
Just about to head on vacation, so I will be able to make these updates in a few weeks.

david-mohr:
Is there a way I can run these github actions on my own account? I have not used github actions before, so not familiar with them. Plus I seem to be in a different timezone making it challenging to work on it while you are available.

david-mohr:
Thanks for the pointers, the workflow should now be working as expected.

icappello:
We'll give you a feedback as soon as possible.

Thanks!

src/client.ts
```diff
@@ -707,6 +707,7 @@ export interface ClientParams {
   port: number
   username: string
   password: string
+  mechanism?: string
```
gpad, line 710: Instead of declaring as generic string why don't declaring as:
```suggestion
  mechanism?: "PLAIN" | "EXTERNAL"
```

src/requests/sasl_authenticate_request.ts
```diff
@@ -14,10 +14,19 @@ export class SaslAuthenticateRequest extends AbstractRequest {
...
+        writer.writeUInt8(0)
+        writer.writeData(this.params.username)
+        writer.writeUInt8(0)
+        writer.writeData(this.params.password)
+        break
+      case "EXTERNAL":
+        writer.writeUInt32(0)
+        break
+      default:
+        throw new Error(`Auth mechanism ${this.params.mechanism} not implemented`)
```
gpad, line 29:
```suggestion
      default:
      assertUnreachable(this.params.mechanism)
      
      ... 
      function(mechanism: never) {
        throw new Error(`Auth mechanism ${mechanism} not implemented`)
        }
```

david-mohr:
@gpad thanks, updates applied

icappello (approved):

david-mohr:
@icappello just curious, do you have an eta on the next release? Happy using my fork for now, but wanted to follow up before I forget. Thanks!

PR #218, opened 2024-12-02
217 fix super stream consumer property

### LINKED ISSUE #217: Single active consumer on Super Stream does not balance consumers across partitions
I've been comparing my application to the Java-based reference example (https://github.com/acogoluegnes/rabbitmq-stream-single-active-consumer#super-streams) and I'm seeing two issues with my current super-stream / single-active-consumer setup which **makes me think there is something missing/wrong in the way I have configured my consumers (as described in my comment in this other issue https://github.com/coders51/rabbitmq-stream-js-client/issues/216#issuecomment-2483242386)**:
1.  `rabbit-streams list_stream_consumer_groups` is showing `-1` for the `partition_index` unlike the Java implementation and according to the doc that means the streams are not part of a super stream
2. When I start a new instance of my application the single active connections are not rebalanced as they are in the Java example.

I'm not sure if these are related, but at the moment this means I do not get any scaling - the messages are distributed across the streams of the super stream, but all the messages go to the same client.

Notes:
- I've tried creating the superstream using the JS client library and the `rabbit-streams` command and both seem to produce the same topology. The attributes on the superstream direct exchange and the bindings to the streams look correct (the binding routing key and `x-stream-partition-order` arguments map to `0`,`1`,`2`, etc as expected) and seem to match those in the Java example.
- Version information: `RabbitMQ 3.13.7`  `Erlang 26.2.5.5`
- I have a suspicion that the `super-stream` consumer property might be involved in the `partition_index` being `-1` since I see evidence that the individual stream consumers created in the Java client have this property set to the super stream name, but I couldn't find reference to it in the JS client.

_Originally posted by @tunniclm in https://github.com/coders51/rabbitmq-stream-js-client/issues/216#issuecomment-2483762662_

tunniclm:
Fixes #217 

---

The active part of the change is adding the property to the consumer:
https://github.com/coders51/rabbitmq-stream-js-client/pull/218/files#diff-25d66d74617fe2e23d7946bd6e3ba95640ab1b9bc8947445d604fc271c7c1f12R490-R492
```ts
    if (superStream) {
      properties["super-stream"] = superStream
    }
```

The rest is plumbing to get the super stream name passed down to the right place.

I have added a test which fails on `main` and passes on this branch. I have included it inside the existing super stream consumer end-to-end test.

I'd be happy to work through any comments or suggested restructures in order to get this in shape to be merged.

Thanks

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -184,7 +184,11 @@ export class Client {
     return res.ok
   }
 
-  public async declareConsumer(params: DeclareConsumerParams, handle: ConsumerFunc): Promise<Consumer> {
+  public async declareConsumer(
+    params: DeclareConsumerParams,
+    handle: ConsumerFunc,
+    superStreamConsumer?: SuperStreamConsumer
+  ): Promise<Consumer> {
     const connection = await this.getConnection(params.stream, "consumer", params.connectionClosedListener)
     const consumerId = connection.getNextConsumerId()
 
@@ -211,7 +215,7 @@ export class Client {
       await this.closeConsumer(consumer.extendedId)
     })
     this.consumers.set(consumer.extendedId, { connection, consumer, params })
-    await this.declareConsumerOnConnection(params, consumerId, connection)
+    await this.declareConsumerOnConnection(params, consumerId, connection, superStreamConsumer?.superStream)
     this.logger.info(
       `New consumer created with stream name ${params.stream}, consumer id ${consumerId} and offset ${params.offset.type}`
     )
@@ -245,6 +249,7 @@ export class Client {
   ): Promise<SuperStreamConsumer> {
     const partitions = await this.queryPartitions({ superStream })
     return SuperStreamConsumer.create(handle, {
+      superStream,
       locator: this,
       consumerRef: consumerRef || `${superStream}-${randomUUID()}`,
       offset: offset || Offset.first(),
@@ -468,7 +473,12 @@ export class Client {
     }
   }
 
-  private async declareConsumerOnConnection(params: DeclareConsumerParams, consumerId: number, connection: Connection) {
+  private async declareConsumerOnConnection(
+    params: DeclareConsumerParams,
+    consumerId: number,
+    connection: Connection,
+    superStream?: string
+  ) {
     const properties: Record<string, string> = {}
     if (params.singleActive && !params.consumerRef) {
       throw new Error("consumerRef is mandatory when declaring a single active consumer")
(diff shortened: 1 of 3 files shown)


icappello:
Hello, thanks for your PR.
We will give you a feedback as soon as possible.

Gsantomaggio:
@tunniclm Thank you. 
Yes, the fix is needed to enable SaC with super stream

src/client.ts
```diff
@@ -184,7 +184,11 @@ export class Client {
     return res.ok
   }
 
-  public async declareConsumer(params: DeclareConsumerParams, handle: ConsumerFunc): Promise<Consumer> {
+  public async declareConsumer(
+    params: DeclareConsumerParams,
+    handle: ConsumerFunc,
+    superStreamConsumer?: SuperStreamConsumer
```
tarzacodes, line 190: Couldn't we simply pass the name contained in the `superStream` property  instead of the whole object?

src/super_stream_consumer.ts
```diff
@@ -29,7 +32,8 @@ export class SuperStreamConsumer {
       this.partitions.map(async (p) => {
         const partitionConsumer = await this.locator.declareConsumer(
           { stream: p, consumerRef: this.consumerRef, offset: this.offset, singleActive: true },
-          this.handle
+          this.handle,
+          this
```
tarzacodes, line 36: Relevant to the other comment - here we could be passing just the `this.superStream`

tarzacodes:
Brilliant, very good work and thanks! Added just a very minor comment, but everything else looks spiffy and clean :sparkles:

src/client.ts
tunniclm: Yes, for sure. The concern I had was that declareConsumer() is a public-facing api and there's not much hint that you shouldn't pass in a super steam name if you don't have a legitimate super stream.

My first thought was to add an internal version of `declareConsumer()` that would be called by `SuperStreamConsumer` only and not be publically accessible to make incorrect usage impossible - however, it's a bit tricky to do and I didn't like the complexity it added.

Instead I chose this simpler way of indicating you should have a valid super stream to call it that way.

This is one of the choices I was hoping for a steer on, since I'm not sure what the best fit would be for this project. If you prefer `superStreamName: string` here, I'd be happy to make that change.

tarzacodes: Your idea is actually very valid, I did not think of that. I'm still not a superfan of adding a different object to the declareConsumer interface - even as an optional. But I started trying out coding a solution to this and I quickly reached the same conclusion you did: this would add a significant amount of extra complexity and at this stage it is simply not worth it. So I can only agree with your findings, good work!

tarzacodes:
LGTM for me, just needs a rebase! I haven't worked on this project for a minute now so maybe a second pair of eyes might be ideal.

tunniclm:
Rebased on latest `main`.

dc1992:
@tunniclm the merge has been blocked due to the CI failing, can you fix that?

dc1992 (approved):

PR #224, opened 2025-01-13
214 question async consumerfunc

dc1992:


test/e2e/address_resolver.test.ts
```diff
@@ -40,11 +40,15 @@ describe("address resolver", () => {
   })
 
   it("declaring a consumer - should not throw", async () => {
-    await client.declareConsumer({ stream: streamName, offset: Offset.first() }, () => null)
+    await client.declareConsumer({ stream: streamName, offset: Offset.first() }, async () => {
+      return
+    })
```
tarzacodes, line 45: Maybe add a test for the sync consumer function - we support both right?

tarzacodes (changes requested):
The `ConsumerFunc` can be async or sync if I'm not mistaken - a lot of these changes, in the tests, are to substitute sync functions with async ones which makes the PR a lot harder to digest and check. Unless those changes in the tests are necessary to highlight some change in behaviour, I'd revert them

test/e2e/close_consumer.test.ts
```diff
@@ -38,7 +38,9 @@ describe("close consumer", () => {
 
   it("closing a consumer in an existing stream", async () => {
     await client.declarePublisher({ stream: testStreamName })
-    const consumer = await client.declareConsumer({ stream: testStreamName, offset: Offset.first() }, () => null)
+    const consumer = await client.declareConsumer({ stream: testStreamName, offset: Offset.first() }, async () => {
+      return
+    })
```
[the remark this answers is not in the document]
dc1992: This is actually needed because typescript complains "Type 'null' is not assignable to type 'void | Promise<void>'". It can either be sync or async, but it has to do or return something in order to not have typescript complaining

test/e2e/address_resolver.test.ts
dc1992: I can remove `async` in few tests I modified to cover both cases, but I wouldn't add more tests since it's not even a different case. What do you think? Does it make sense to you?

tarzacodes: Tests are also documentation - might be that this is not the right place, but I would have two different tests that test the fact that you can declare a consumer with two types of ConsumerFunc. Probably in `declare_consumer.test.ts` we might duplicate the `declaring a consumer on an existing stream - the consumer should handle the message` test and have it in two versions, with sync or async ConsumerFunc

test/e2e/close_consumer.test.ts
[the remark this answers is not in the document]
tarzacodes: Ah, that makes sense. But we could still remove the async from the signature right?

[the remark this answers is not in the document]
dc1992: Yes exactly, the signature can either be async or not

test/e2e/address_resolver.test.ts
dc1992: Added the test you suggested

tarzacodes (approved):

icappello (approved):

PR #229, opened 2025-01-21
228 some tests are flaky

dc1992:


.github/workflows/main.yml
```diff
@@ -20,7 +20,7 @@ jobs:
 
     services:
       rabbitmq:
-        image: rabbitmq:3.13-rc-management
```
tarzacodes, line 23: Is there any test that was flaky specifically because of the version? Some bug that got fixed?

dc1992: No, I updated it just because that's the latest stable version

test/e2e/offset.test.ts
```diff
@@ -163,8 +164,8 @@ describe("offset", () => {
 
       await eventually(async () => {
         expect(receivedMessages).to.have.length(laterMessages.length)
-      })
-    })
+      }, 2500)
+    }).timeout(7000)
```
icappello, line 168: I mention this since we are focusing on improving tests in this PR: after we `eventually` get the expected number of messages, we could also check that we `always` have that amount (the utility function is already defined in `test/support/util.ts`)

test/e2e/offset.test.ts
```diff
@@ -150,8 +150,9 @@ describe("offset", () => {
       const receivedMessages: Message[] = []
       const publisher = await client.declarePublisher({ stream: testStreamName })
       const previousMessages = await sendANumberOfRandomMessages(publisher)
-      await wait(10)
+      await wait(1000)
       const offset = new Date()
+      await wait(1000)
```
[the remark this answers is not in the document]
dc1992: Done in the way you suggested

test/e2e/offset.test.ts
dc1992: Added `always` check

.github/workflows/main.yml
tarzacodes: As a rule of thumb, it might be better to leave version bumps to separate MRs, especially if it's a major version bump, but I don't want to leave this MR hanging for nitpicking, so resolving the conversation and approving this!

tarzacodes (approved):

PR #233, opened 2025-01-30
230 change stream options when already created

dc1992:
In this PR I removed the `boot` parameter that was used to activate the deduplication. From now, to activate the deduplication you just need to define a `publisherRef` value.

I also fixed a race condition when sending multiple messages concurrently with deduplication actived.


Gsantomaggio (changes requested):
To use the deduplication, the client needs to expose an API to pass the `publishingid`; otherwise it does not make sense :)

See for example the Go-client:
https://github.com/rabbitmq/rabbitmq-stream-go-client/blob/main/examples/deduplication/deduplication.go#L66-L84
where the user _must_ provide the id for each message to avoid duplications 

or the .NET client where with the message you can pass the id the API
https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/main/RabbitMQ.Stream.Client/Reliable/DeduplicatingProducer.cs#L71

dc1992:
> To use the deduplication, the client needs to expose an API to pass the `publishingid`; otherwise it does not make sense :)
> 
> See for example the Go-client: https://github.com/rabbitmq/rabbitmq-stream-go-client/blob/main/examples/deduplication/deduplication.go#L66-L84 where the user _must_ provide the id for each message to avoid duplications
> 
> or the .NET client where with the message you can pass the id the API https://github.com/rabbitmq/rabbitmq-stream-dotnet-client/blob/main/RabbitMQ.Stream.Client/Reliable/DeduplicatingProducer.cs#L71

Ok, but passing `publishingid` from outside is not mandatory, is it?
If you don't pass it from outside the client takes the next id increasing its own by one, right?

Gsantomaggio:
> Ok, but passing publishingid from outside is not mandatory, is it?

It is mandatory else the deduplication won't work. In this implementation, deduplication works like the standard producer.

README.md
```diff
@@ -123,6 +123,39 @@ await publisher.send(Buffer.from("my message content"))
 await client.close()
 ```
 
+### Deduplication
+
+If you want to make sure that a given message isn't sent more than once, you must use deduplication.
+To create a publisher with deduplication, you just need to pass `publisherRef` to the `declarePublisher` function, this way RabbitMQ will detect messages with lower ids and discard them.
+Note that these ids are incremental, so you have to be careful about how your application publishes the messages, as the order cannot be guaranteed in multi-threaded applications.
+It's also important to remember that the client doesn't control that the `publisherRef` is unique, it's the user's responsibility to guarantee that.
```
Gsantomaggio, line 131: Note:
The server does not control the `publisherRef` across the producers. It's the user's responsibility to guarantee that.

example/src/deduplication_example.js
```diff
@@ -0,0 +1,64 @@
...
+  await firstDeduplicationPublisher.send(Buffer.from("Test message 4"), { publishingId: 1n }) //this message won't be sent
+  await firstDeduplicationPublisher.flush()
+  const firstPublisherPublishingId = await firstDeduplicationPublisher.getLastPublishingId()
+  await firstDeduplicationPublisher.close()
+
+  console.log(`Publishing id is ${firstPublisherPublishingId} (must be 3)`) //this must be the greatest publishingId sent, 3 in this case
+
+  const secondDeduplicationPublisher = await client.declarePublisher({ stream: streamName, publisherRef: publisherRef })
+  //with the second publisher if we try to send messages with lower publishingId than the last one, they will not be sent
+  await secondDeduplicationPublisher.send(Buffer.from("Test message 5"), { publishingId: 1n }) //won't be sent
```
Gsantomaggio, line 44:
```
will not be stored
```

example/src/deduplication_example.js
```diff
@@ -0,0 +1,64 @@
...
+  await secondDeduplicationPublisher.send(Buffer.from("Test message 6"), { publishingId: 2n }) //won't be sent
+  await secondDeduplicationPublisher.send(Buffer.from("Test message 7"), { publishingId: 7n }) //this will be sent since 7 is greater than 3, the last highest publishingId
```
Gsantomaggio, line 46: will be stored

Gsantomaggio (changes requested):

example/src/deduplication_example.js
```diff
@@ -0,0 +1,64 @@
...
+  //to declare a publisher with deduplication enabled, you need to set a publisherRef
+  const firstDeduplicationPublisher = await client.declarePublisher({ stream: streamName, publisherRef: publisherRef })
+
+  //with deduplication actived, you can send messages without a publishingId; in this case it will be incremental
+  await firstDeduplicationPublisher.send(Buffer.from("Test message 1")) //publishingId = 1
+  await firstDeduplicationPublisher.send(Buffer.from("Test message 2")) //publishingId = 2
+  //but you can also set a publishingId, note that it must be greater than the last one for the message to be sent
+  await firstDeduplicationPublisher.send(Buffer.from("Test message 3"), { publishingId: 3n }) //publishingId = 3
+  //if you choose a publishingId that is less than the last one, the message will not be sent
+  await firstDeduplicationPublisher.send(Buffer.from("Test message 4"), { publishingId: 1n }) //this message won't be sent
```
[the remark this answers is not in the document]
dc1992: Replaced `sent` with `stored`

example/src/deduplication_example.js
dc1992: Replaced `sent` with `stored`

example/src/deduplication_example.js
dc1992: Replaced `sent` with `stored`

README.md
dc1992: Are you suggesting to replace this
>It's also important to remember that the client doesn't control that the `publisherRef` is unique, it's the user's responsibility to guarantee that.

with this?

>Note: The server does not control the publisherRef across the producers. It's the user's responsibility to guarantee that.

Gsantomaggio: yes

dc1992: Ok replaced. Let me know if there is anything else

Gsantomaggio (approved):
Thank you

icappello (approved):

PR #241, opened 2025-03-14
fix: get default vhost if empty string is provided

l4mby:


spirin22:
I would prefer to get a validation error rather than an implicit fallback.

dc1992 (changes requested):

dc1992 (approved):

PR #244, opened 2025-03-24
fix: handle consumer unsubscribe

l4mby:


src/client.ts
```diff
@@ -703,6 +702,27 @@ export class Client {
...
+
+  private async closing(consumer: StreamConsumer, extendedConsumerId: string) {
+    await consumer.close(true)
+    this.consumers.delete(extendedConsumerId)
+    this.logger.info(`Closed consumer with id: ${extendedConsumerId}`)
+  }
+
+  private streamExistsGiven(streamInfo: StreamMetadata) {
+    return (
+      streamInfo.responseCode !== ResponseCode.StreamDoesNotExist &&
```
dc1992, line 721: `private streamExistsGiven(streamInfo: StreamMetadata) => streamInfo.responseCode === ResponseCode.SubscriptionIdDoesNotExist`

this should be enough: if responseCode must be equal to ResponseCode.SubscriptionIdDoesNotExist, you don't need to check that it is not equal to ResponseCode.StreamDoesNotExis

dc1992 (changes requested):

src/client.ts
l4mby: sorry i forgot an exclamation mark, basically here i am checking that the stream and consumer still exist so if the response code is either SubscriptionIdDoesNotExist or StreamDoesNotExist i should not run the unsubscribe

dc1992 (dismissed):

PR #246, opened 2025-04-09
Fix bug when connecting to the same stream name different vhosts

david-mohr:
During testing, we ran into a problem where we would create two connections to different vhosts, each with the same stream name. We could see in the RabbitMQ management interface that the configuration is valid and accessing them individually would work as expected, but when trying to access them at the same time, we would get messages leaking into the wrong vhost.

After debugging the issue, I tracked the root cause to the `ConnectionPool.getCacheKey` function which would build a cache using only the stream name, not the vhost. This meant that we were always getting the cached copy regardless of the actual vhost.

All the other changes are to support that one change and there is a test to confirm the result.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -486,7 +486,7 @@ export class Connection {
   }
 
   private virtualHostIsNotValid(virtualHost: string) {
-    if (!virtualHost || virtualHost.split("/").length !== 2) {
+    if (!virtualHost) {
       return true
     }
 
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -62,6 +62,7 @@ export type ConnectionInfo = {
   port: number
   id: string
   ready: boolean
+  vhost: string
   readable?: boolean
   writable?: boolean
   localPort?: number
@@ -376,6 +377,7 @@ export class Connection {
       writable: this.socket.writable,
       localPort: this.socket.localPort,
       ready: this.ready,
+      vhost: this.vhost
     }
   }
 
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -78,6 +78,7 @@ type ListenerEntry = {
 
 export class Connection {
   public readonly hostname: string
+  public readonly vhost: string
   public readonly leader: boolean
   public readonly streamName: string | undefined
   private socket: Socket
@@ -109,6 +110,7 @@ export class Connection {
     private readonly logger: Logger
   ) {
     this.hostname = params.hostname
+    this.vhost = params.vhost
     this.leader = params.leader ?? false
     this.streamName = params.streamName
     if (params.frameMax) this.frameMax = params.frameMax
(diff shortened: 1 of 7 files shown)


david-mohr:
The failing test is also broken on `main`, so I have ignored it

l4mby:
Hello, thanks a lot for taking the time to open the pr. We will check it and will comment it in the following days.
Yes we are leaving that test on purpose, so it can be ignored.
Thank you again.

src/connection.ts
```diff
@@ -484,7 +488,7 @@ export class Connection {
   }
 
   private virtualHostIsNotValid(virtualHost: string) {
-    if (!virtualHost || virtualHost.split("/").length !== 2) {
```
l4mby, line 487: Hi again, this check is needed for the validity of the vhost.
It's something similar to what the java client does, for example.
(https://github.com/rabbitmq/rabbitmq-stream-java-client/blob/main/src/main/java/com/rabbitmq/stream/impl/StreamEnvironment.java#L459)

david-mohr: The code you linked to is for pulling the vhost out of a URI and then decoding it (e.g. the vhost would be `/%2F` for the default vhost). The vhost arriving at the JS function is already a decoded string, so can take the form of any valid vhost name (and I cannot find any restrictions on hostnames)

l4mby: yes it is already a decoded string, but it can be any string you like, we could put a regex instead.

david-mohr: regex would be fine. What pattern would you propose? I can't seem to find much in the official docs

l4mby: you can use this pattern maybe `/^\/\w+$/g` it's still not perfect but good enough

david-mohr: I tried this and got
```
Error: [ERROR]: VirtualHost '/' is not valid
```
So I made the "word" portion optional `/^\/(\w+)?$/g` and now I get
```
Error: [ERROR]: VirtualHost 'vhost1' is not valid
```
Since it doesn't start with forward slash.

I'm still quite confident that we don't need this check at this point in the code. I did some testing and even things like `///` and `@!` are still valid vhost names. To use them with the equivalent Java function, they would need to be URI encoded and made into a path, e.g. `/%2F%2F%2F` and `/%40!`

david-mohr: I also ran a quick test on some of our existing code and the current `main` branch breaks our working code saying that the vhost names are invalid

l4mby (dismissed):

l4mby:
@david-mohr can you please rebase your branch? we fixed the failing pipeline, so it should be correct now.

david-mohr:
I ran the tests locally using node v20 and they passed, so this failing test is unexpected...

david-mohr:
@l4mby I tried adding a minor change to re-trigger the CI tests and now a different test is failing. I don't think these are related to my change and running the tests locally everything still passes. Not sure how to proceed from here... any advice is welcome.

l4mby (approved):

PR #264, opened 2025-06-13
260 refactoring: manage connections inside pool

apietroni51:
So far we kept the `manuallyClose` param, but the `refCount` and connection caching are managed inside the `ConnectionPool`.

### DIFF
--- a/src/consumer.ts
+++ b/src/consumer.ts
@@ -97,12 +97,9 @@ export class StreamConsumer implements Consumer {
     this.singleActive = params.singleActive ?? false
   }
 
-  async close(manuallyClose: boolean): Promise<void> {
+  async close(): Promise<void> {
     this.closed = true
-    this.connection.decrRefCount()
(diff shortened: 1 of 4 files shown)


src/consumer.ts
```diff
@@ -97,12 +97,9 @@ export class StreamConsumer implements Consumer {
...
   }
 
-  async close(manuallyClose: boolean): Promise<void> {
+  async close(): Promise<void> {
     this.closed = true
-    this.connection.decrRefCount()
-    if (this.pool.removeIfUnused(this.connection)) {
-      await this.connection.close({ closingCode: 0, closingReason: "", manuallyClose })
-    }
+    await this.pool.releaseConnection(this.connection)
```
tarzacodes, line 102: At this point I'm having a hard time understanding why the consumer always releases the connection "not manually" - even though this method *is* still public so it could be the end user that invokes it. Would someone be able to clarify this, even by pointing to specific tests?

albertobarrila (approved):

Issue #265, opened 2025-06-16
As a user, I want to avoid specifying all the options for enabling TLS


apietroni51:
In `ClientParams`, `ssl` should be optional, or boolean as a shortcut. All `SSLConnectionParams` fields should be optional as supported by `tls.connect()`

```typescript
export interface SSLConnectionParams {
  key?: string
  cert?: string
  ca?: string
  rejectUnauthorized?: boolean
}
```

Pseudocode for the socket creation:
```typescript
private createSocket() {
    const socket = this.params.ssl
      ? tls.connect(this.params.port, this.params.hostname, typeof this.params.ssl === Boolean ? {} : this.params.ssl)
      : new Socket().connect(this.params.port, this.params.hostname)
    if (this.params.socketTimeout) socket.setTimeout(this.params.socketTimeout)
    return socket
}
```

This will also address https://github.com/coders51/rabbitmq-stream-js-client/issues/254

PR #266, opened 2025-06-16
#265 optional ssl params, wip on rejectUnauthorized

apietroni51:


apietroni51:
@tarzacodes @icappello @albertobarrila
we have made the SSL params optional reflecting tls.connect() but we're not sure what to do with `rejectUnauthorized`, which was currently overridden to `false`. What should we do with it?

PR #269, opened 2025-06-17
262# hide manuallyClose param in close methods

apietroni51:
As suggested by @tarzacodes we have wrapped the original StreamConsumer and StreamPublisher, that expose the `manuallyClose` parameter for internal reasons (i.e. callbacks), and created a new interface for each that just expose a parameterless `close()`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x, 20.x, 22.x]
+        node-version: [18.x, 20.x, 22.x, 24.x]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
 
     services:
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -1,5 +1,5 @@
 module.exports = {
-  require: ["ts-node/register/transpile-only", "test/index.ts"],
+  require: ["tsx", "test/index.ts"],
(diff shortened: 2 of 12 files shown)


albertobarrila:
The .close(false) is called only in the `registerForClosePublisher` of the client and only for the `Publisher` otherwise is passed true (that is the default of the param)

I honestly do not like the solution of having a `Private` that shadow the entire class only to hide a single parameter, maybe is "better" to have a special method to force that param to be called only in that specific case...

something like

```
// standard method to close the connection
public async close(): Promise<void> {
    if (!this.closed) {
        await this.flush()
        await this.pool.releaseConnection(this.connection)
    }
    this._closed = true
}

// special case where to force that param
// maybe find a better name I'm very bad at it
public async forceClose(): Promise<void> {
    if (!this.closed) {
        await this.flush()
        await this.pool.releaseConnection(this.connection, false)
    }
    this._closed = true
}
```
@gpad wdyt? @tarzacodes maybe you have already tried this route and found something that I've lost....

albertobarrila (approved):
this first solution is a first step towards a simplification (there's more wax to remove) but is ok for now.
I've also swapped ts-node for tsx to support also node 24

albertobarrila (approved):
this first solution is a first step towards a simplification (there's more wax to remove) but is ok for now.
I've also swapped ts-node for tsx to support also node 24

PR #282, opened 2025-10-29
fix: reading bytes in log

l4mby:
We remove the call to `readInt32BE` since the length of bytes we receive is variable


sieniahin:
Thanks for fixing this! 
Do you have any plans on releasing it?

PR #285, opened 2025-11-12
feat: queryOffset to return RMQ response code in the domain exception (resolves #283)

### LINKED ISSUE #283: [Question] Query offset command returned error with code 19
Hello, I try to understand if that exception should be actually treated as the case "query offset is not set" or not.

That comes from the line
```
const offset = await client.queryOffset({ reference: consumerRef, stream: streamName })
```

when I try to get the stream offset.

WhereJuly:
Hey Team,

In response to #283 issue. 

- **Before**: the `src\connection.ts->Connection.queryOffset` threw the generic Error with a string variable for any RabbitMQ stream protocol response code including NO_OFFSET (19). See more in #283.

- **After**:  the `src\connection.ts->Connection.queryOffset` throws the domain exception `src\application\Code51Exception.ts` with a dedicated code field. See the usage example in the class JSDoc block.

The main changes introduced:

- The domain exception `src\application\Code51Exception.ts` to carry the response code along the error message, tested with `test\unit\application\Code51Exception.test.ts`
- A `NoOffset` member into`src\util.ts->ResponseCode`
- A `src\application\Code51Exception.ts->TResponseCode` type alias to simplify the response code parameters description.

A number of JSDoc blocks in:

- `src\connection.ts`: `storeOffset`, `queryOffset`
- `src\util.ts`: `ResponseCode`
- `src\client.ts`: `DeclareConsumerParams`, `StoreOffsetParams`, `QueryOffsetParams`

The CI attempt went well.

Thanks.

### DIFF
--- a/src/application/Code51Exception.ts
+++ b/src/application/Code51Exception.ts
@@ -1,0 +1,25 @@
+"use strict"
+
+import { ResponseCode } from "../util"
+
+type TResponseCode = (typeof ResponseCode)[keyof typeof ResponseCode]
+
+export default class Code51Exception extends Error {
+  readonly #code?: TResponseCode
+
+  constructor(message: string, rmqStreamResponseCode?: TResponseCode) {
+    super(message)
+
+    this.name = this.constructor.name
+    this.#code = rmqStreamResponseCode ?? undefined
+
+    // Maintains proper stack trace for where our error was thrown (only available on V8)
+    if (Error.captureStackTrace) {
+      Error.captureStackTrace(this, this.constructor)
+    }
+  }
+
+  public get code(): TResponseCode | undefined {
+    return this.#code
+  }
+}
--- a/src/application/Code51Exception.ts
+++ b/src/application/Code51Exception.ts
@@ -2,7 +2,7 @@
 
 import { ResponseCode } from "../util"
 
-type TResponseCode = (typeof ResponseCode)[keyof typeof ResponseCode]
+export type TResponseCode = (typeof ResponseCode)[keyof typeof ResponseCode]
 
 export default class Code51Exception extends Error {
   readonly #code?: TResponseCode
--- a/src/application/Code51Exception.ts
+++ b/src/application/Code51Exception.ts
@@ -10,6 +10,8 @@ export default class Code51Exception extends Error {
   constructor(message: string, rmqStreamResponseCode?: TResponseCode) {
     super(message)
 
+    Object.setPrototypeOf(this, new.target.prototype)
+
     this.name = this.constructor.name
     this.#code = rmqStreamResponseCode ?? undefined
 
--- a/src/application/Code51Exception.ts
+++ b/src/application/Code51Exception.ts
@@ -4,6 +4,41 @@ import { ResponseCode } from "../util"
 
 export type TResponseCode = (typeof ResponseCode)[keyof typeof ResponseCode]
 
+/**
+ * Provides distinct domain exception for the package. Contains the optional
+ * RabbitMQ Stream protocol response code for more convenient processing.
+ *
+ * @param message A custom error message.
+ * @param rmqStreamResponseCode The above mentioned response code.
+ *
+ * @see https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_stream/docs/PROTOCOL.adoc#response-codes
+ *
+ * @example Selectively manage the exception type and react differently.
+ *
+ * ```typescript
+ * let result: any;
+ *
+ * const isRethrowable = (error_: Error) => {
+ *     const isGenericError = error_ instanceof Code51Exception;
+ *     const isNonManagedResponseCode = (error_ as Code51Exception).code !== ResponseCode.NoOffset;
+ *
+ *     return isGenericError && isNonManagedResponseCode;
+ * };
+ *
+ * try {
+ *     result = consumer.queryOffset();
+ *     // ... process result
+ * } catch (error_) {
+ *     if (isRethrowable(error_)) { throw error_; }
+ *
+ *     const error = error_ as Code51Exception;
+ *     if (error.code === ResponseCode.NoOffset) { return null; }
+ *
+ *     return result;
+ * }
+ * ```
+ *
+ */
 export default class Code51Exception extends Error {
   readonly #code?: TResponseCode
 
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -565,6 +565,37 @@ export class Connection {
     return this.send(new StoreOffsetRequest(params))
   }
 
+  /**
+   * Return the server-side saved offset.
+   *
+   * @see https://www.rabbitmq.com/tutorials/tutorial-two-javascript-stream
+   * @see https://www.rabbitmq.com/blog/2021/09/13/rabbitmq-streams-offset-tracking
+   *
+   * @params QueryOffsetParams Stream name and the named consumer identifier.
+   *
+   * @see {@link storeOffset}
+   * @see {@link Client.connect}
+   *
+   * @example
+   *
+   * Consumer reads previously saved offset to start consumning from the desired offset.
+   * Note, to be server-tracked the offset shuould be explicitly saved with `client.storeOffset()`
+   * On consumer side create the client, detect the server-side saved offset
+   * to detect the deisred offset to consume from. Then declare a consumer with the desired offset.
+   * When consuming, the consumer message handler above saved the offset server-side to be
+   * able to further track it
+   *
+   * ```typescript
+   * // ... create client
+   * // Detect the desider starting offset for the next stream operation.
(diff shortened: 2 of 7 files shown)


src/application/Code51Exception.ts
```diff
@@ -0,0 +1,62 @@
+"use strict"
+
+import { ResponseCode } from "../util"
+
+export type TResponseCode = (typeof ResponseCode)[keyof typeof ResponseCode]
```
l4mby: I don't think this type is necessary, we can just use a type number for the code

src/application/Code51Exception.ts
```diff
@@ -0,0 +1,62 @@
+"use strict"
```
l4mby: Can you follow our style of code? For example:
- no "use strict";
- the name of the files should be in snake_case;

src/application/Code51Exception.ts
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @param message A custom error message.
+ * @param rmqStreamResponseCode The above mentioned response code.
+ *
+ * @see https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_stream/docs/PROTOCOL.adoc#response-codes
+ *
+ * @example Selectively manage the exception type and react differently.
+ *
+ * ```typescript
+ * let result: any;
```
l4mby: The result here should be a `Promise<bigint>` if not awaited

src/application/Code51Exception.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    Object.setPrototypeOf(this, new.target.prototype)
+
+    this.name = this.constructor.name
+    this.#code = rmqStreamResponseCode ?? undefined
+
+    // Maintains proper stack trace for where our error was thrown (only available on V8)
+    if (Error.captureStackTrace) {
+      Error.captureStackTrace(this, this.constructor)
+    }
```
l4mby: this case here we can remove it

src/connection.ts
```diff
@@ -558,15 +559,70 @@ export class Connection {
...
+   * @see https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_stream/docs/PROTOCOL.adoc#response-codes
+   */
   public async queryOffset(params: QueryOffsetParams): Promise<bigint> {
     this.logger.debug(`Query Offset...`)
     const res = await this.sendAndWait<QueryOffsetResponse>(new QueryOffsetRequest(params))
     if (!res.ok) {
-      throw new Error(`Query offset command returned error with code ${res.code}`)
+      const code = res.code as TResponseCode
+
+      throw new Code51Exception(`Query offset command returned error with code ${res.code}`, code)
```
l4mby: Here since we will use the code as a number we can remove the as and throw directly the code in the constructor of `QueryOffsetError`

test/e2e/offset.test.ts
```diff
@@ -292,5 +294,24 @@ describe("offset", () => {
       await wait(200)
       await expectToThrowAsync(() => consumer.queryOffset(), Error, `This socket has been ended by the other party`)
     })
+
+    it("query offset is able to raise Code51Exception with ResponseCode.NoOffset code value set if there is no offset", async () => {
+      const params = { stream: testStreamName, consumerRef: "my_consumer", offset: Offset.first() }
```
l4mby: Here we can follow the same style of code for the tests, for example the handler declared as parameter directly inside the declareConsumer function, test format in AAA

Instead of throwing the exception you can just catch the error and casting it to `QueryOffsetError` after the second instanceOf check

l4mby (commented):
We reviewed your code and found some inconsistencies with our style. Please fix it so we can proceed with the integration. Thank you.
PS: Did you use any AI tool? If yes can you please tell us which tool/model you have used?

l4mby:
Hi @WhereJuly thank you for the infos. Yes we talked about it internally and decided it's ok to keep the JSDoc in the PR.
Thank you again in advance

src/application/Code51Exception.ts
```diff
@@ -0,0 +1,62 @@
...
+ *     const error = error_ as Code51Exception;
+ *     if (error.code === ResponseCode.NoOffset) { return null; }
+ *
+ *     return result;
+ * }
+ * ```
+ *
+ */
+export default class Code51Exception extends Error {
+  readonly #code?: TResponseCode
```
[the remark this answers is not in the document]
WhereJuly: @l4mby  Thank you. Understood. Wanted to discuss the following points.

>  `code`: make it mandatory

Will do.

> more appropriate name for the error, specific to the queryoffset, for example `QueryOffsetError`

I planned this error to be a generic error for the entire package. Now I see this seems too wide. On the other hand maybe `QueryOffsetError` is too specific? 

Could it be better `RMQProtocolResponseError` or something like that to cater for all potential use cases with the protocol response code? To not end up with a number of several too specific error classes?

>  Also can you rename the property in code

I could but this is not name convention, `#` is the JS property visiblity level. It actually makes the JS (not TS) [property private](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements) at runtime.

The reason to have it is to actually make it read-only at run time, accessibe only via a getter. Making it `code` we expose it at runtime. Do you still think we should not use JS private properties?

src/application/Code51Exception.ts
WhereJuly: @l4mby I see.

```typescript
export default class RMQProtocolResponseError extends Error {
  readonly #code: TResponseCode

  constructor(message: string, rmqStreamResponseCode: TResponseCode) {
    super(message)
```

The retionale to narrow type of `rmqStreamResponseCode` is to not allow to provide any `number` whereas we actually expect only `ResponseCode`. 

The type alias `TResponseCode` is for better clarity and readability - semantically and to not use that long typecast. So I believe we should keep it. Let me know your thought on this.

src/connection.ts
WhereJuly: @l4mby For this case I made an explanation here https://github.com/coders51/rabbitmq-stream-js-client/pull/285#discussion_r2536475272.

Shortly, my logic here is we better use the narrower type `ResponseCode` than wider `number` as the error class is going to accept only `ResponseCode`, so no reason to relax the expected type.

src/application/Code51Exception.ts
gpad: Hi @WhereJuly the problem with this type `TResponseCode` is that force you to use the `as` like in the code below

```typescript
const code = res.code as TResponseCode
```

And if you receive a number that is not present in `ResponseCode` you are forcing it. This is the reason about using a `number` and left the user to verify if it's a known code present in `ResponseCode`.

About the name, it could be useful to have an error we use instead of "generic error" but it not be too generic because not always can be used so I suggest: `class StreamResponseError`.

So this code could became:

```typescript
export class StreamResponseError extends Error {
  constructor(message: string, readonly errorCode: number) {
    super(message)
  }
}
```
If you want to use the `#errorCode` instead the typescript version, it's fine I suggest to name the property `errorCode` or `responseCode`.
As side note we prefer don't use `default` export.

src/connection.ts
gpad: Hi @WhereJuly as written above we try to avoid `as` because it's very dangerous.

gpad (changes requested):

gpad:
> Hey @gpad. I made nearly all updates, the only [comment](https://github.com/coders51/rabbitmq-stream-js-client/pull/285#discussion_r2597240151) is pending your response. Thanks.

Hi @WhereJuly as you can see by the screenshoot below the exception (and the file name) is still `Coders51Exception`. I think you have committed on your branch and the PR is not updated. Do you think it's possibile?
 
image

WhereJuly:
Hey @gpad, yes I did not commit the changes to the upstream. As I wrote, just need your reply to my [comment](https://github.com/coders51/rabbitmq-stream-js-client/pull/285#discussion_r2597240151) above. Then I will proced with submitting the changes. Thank you :)

src/application/Code51Exception.ts
```diff
@@ -0,0 +1,62 @@
...
+ * } catch (error_) {
+ *     if (isRethrowable(error_)) { throw error_; }
+ *
+ *     const error = error_ as Code51Exception;
+ *     if (error.code === ResponseCode.NoOffset) { return null; }
+ *
+ *     return result;
+ * }
+ * ```
+ *
```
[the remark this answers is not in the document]
gpad: About the `@example` you can keep it, obviously you have to update it with the fix we are talking about.
Regarding the `isStreamResponseError` I don't know if a static function can be a guard but IMHO a free function in this context is better you can use it in this way:

```typescript
try {
  ...
} catch (error) {
  if (isStreamResponseError(error)) {
    // here it's a StreamResponseError
  }
```

instead of:

```typescript
try {
  ...
} catch (error) {
  if (StreamResponseError.isStreamResponseError(error)) {
    // here it's a StreamResponseError
  }
```

PR #290, opened 2026-03-16
fix: add missing smalllong (0x55) and long (0x81) AMQP 1.0 format

ThePartyPinguin:
The `decodeFormatCode` function is missing support for the AMQP 1.0 `smalllong` (0x55) and `long` (0x81) type encodings, causing messages containing these types to throw `FormatCode Invalid type 85`.

This occurs in practice when consuming stream messages published by the .NET RabbitMQ client with publisher confirms enabled. The .NET client adds an `x-dotnet-pub-seq-no` header as a `long` value, which the broker encodes as `smalllong` (0x55) for values 0–127. The JS client then fails to parse the message annotations.

Per the [AMQP 1.0 specification](https://www.amqp.org/specification/1.0/amqp-org-download) (section 1.2.5, "List of encodings"), the `long` type has two valid encodings:
- `0x81` — fixed/8, 64-bit two's-complement integer in network byte order
- `0x55` (`smalllong`) — fixed/1, signed long value in the range -128 to 127 inclusive

This PR adds both to the `FormatCode` constants and `decodeFormatCode` switch.

### DIFF
--- a/src/amqp10/decoder.ts
+++ b/src/amqp10/decoder.ts
@@ -31,6 +31,8 @@ export const FormatCode = {
   Int: 0x71,
   SmallUint: 0x52,
   SmallInt: 0x54,
+  SmallLong: 0x55,
+  Long: 0x81,
   Timestamp: 0x83,
   Bool: 0x56,
   BoolTrue: 0x41,
--- a/src/response_decoder.ts
+++ b/src/response_decoder.ts
@@ -438,8 +438,12 @@ export function decodeFormatCode(dataResponse: DataReader, formatCode: number) {
       return dataResponse.readUInt32()
     case FormatCode.SmallInt:
       return dataResponse.readInt8()
+    case FormatCode.SmallLong:
+      return dataResponse.readInt8()
     case FormatCode.Int:
       return dataResponse.readInt32()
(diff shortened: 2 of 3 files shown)


l4mby:
Hi @ThePartyPinguin thank you for opening the pr! We will get back to you in the following days.

gpad:
Hi @ThePartyPinguin do you think it's possible add some test related to this scenario?
If you look at this file: https://github.com/coders51/rabbitmq-stream-js-client/blob/main/test/e2e/declare_consumer.test.ts#L477 we have a consumer that read message with various info inside.

Do you think it's possible add a test related to this scenario?

ThePartyPinguin:
Sorry for the late reply but I've added a new test as you requested;

The new test covers message annotations using the AMQP 1.0 `long` encodings:
  - `0x55` (`smalllong`) for values in the small signed range
  - `0x81` (`long`) for the 64-bit encoding

l4mby:
Hello @ThePartyPinguin thank you again for the support! I looked at the test, what do you think about moving the test in a separate unit test file? since you are testing the parse method for Annotations.
Thank you again for the support!

PR #298, opened 2026-06-09
Add message filtering support for super stream publisher and consumer

SumitAthani:
## Summary

Adds optional message filtering to the super stream publisher and consumer APIs, mirroring the existing filtering support on single-stream `declarePublisher` and `declareConsumer`.

- **`declareSuperStreamPublisher`**: accepts an optional `filter` in params (`FilterFunc`) that is applied to each lazily-created partition publisher for server-side message tagging
- **`declareSuperStreamConsumer`**: accepts an optional `filter` in params (`ConsumerFilter` with `values`, `postFilterFunc`, `matchUnfiltered`) that is applied to each partition consumer

Previously, filtering only worked when calling `declareConsumer` / `declarePublisher` directly on individual partition streams. Super stream wrappers did not expose or forward filter configuration.

**Breaking change:** `declareSuperStreamPublisher` filter was moved from a third argument into the params object to match the consumer API shape.

## Changes

- `src/client.ts` — extend `DeclareSuperStreamPublisherParams` and `DeclareSuperStreamConsumerParams`; forward `filter` in `declareSuperStreamConsumer` / `declareSuperStreamPublisher`
- `src/super_stream_publisher.ts` — thread `FilterFunc` into per-partition `declarePublisher` calls
- `src/super_stream_consumer.ts` — thread `ConsumerFilter` into per-partition `declareConsumer` calls
- `test/e2e/filtering.test.ts` — super stream filtering e2e tests (publish with filter tag, client-side post-filter, server-side bloom filter scenarios)

## Example usage

```typescript
// Publisher: tag messages for server-side filtering
const publisher = await client.declareSuperStreamPublisher(
  {
    superStream: "my-super-stream",
    publisherRef: "my-publisher",
    routingStrategy: "key",
    filter: (msg) => msg.applicationProperties?.["tag"]?.toString(),
  },
  () => routingKey
);

// Consumer: server-side bloom filter + client-side post-filter
await client.declareSuperStreamConsumer(
  {
    superStream: "my-super-stream",
    offset: Offset.first(),
    filter: {
      values: ["A", "B"],
      postFilterFunc: (msg) => msg.applicationProperties?.["tag"] === "A",
      matchUnfiltered: true,
    },
  },
  (msg) => console.log(msg.content.toString())
);

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -380,7 +380,7 @@ export class Client {
   }
 
   public async declareSuperStreamConsumer(
-    { superStream, offset, consumerRef, creditPolicy }: DeclareSuperStreamConsumerParams,
+    { superStream, offset, consumerRef, creditPolicy, filter }: DeclareSuperStreamConsumerParams,
     handle: SuperStreamConsumerFunc
   ): Promise<SuperStreamConsumer> {
     const partitions = await this.queryPartitions({ superStream })
@@ -391,6 +391,7 @@ export class Client {
       offset: offset || Offset.first(),
       partitions,
       creditPolicy,
+      filter,
     })
   }
 
@@ -1065,6 +1066,7 @@ export interface DeclareSuperStreamConsumerParams {
   consumerRef?: string
   offset?: Offset
   creditPolicy?: ConsumerCreditPolicy
+  filter?: ConsumerFilter
 }
 
 export interface SubscribeParams {
--- a/src/client.ts
+++ b/src/client.ts
@@ -394,9 +394,17 @@ export class Client {
     })
   }
 
+  /**
+   * Declare a publisher for a super stream
+   *
+   * @param params - Super stream publisher configuration
+   * @param keyExtractor - Function to extract the routing key from a message
+   * @param filter - Optional filter function for server-side message filtering on each partition publisher
+   */
   public async declareSuperStreamPublisher(
     { superStream, publisherRef, routingStrategy }: DeclareSuperStreamPublisherParams,
-    keyExtractor: MessageKeyExtractorFunction
+    keyExtractor: MessageKeyExtractorFunction,
+    filter?: FilterFunc
   ): Promise<SuperStreamPublisher> {
     return SuperStreamPublisher.create({
       locator: this,
@@ -404,6 +412,7 @@ export class Client {
       keyExtractor,
       publisherRef,
       routingStrategy,
+      filter,
     })
   }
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/client.ts [hunk 1]
```diff
@@ -391,19 +391,29 @@ export class Client {
...
 
+  /**
+   * Declare a publisher for a super stream
+   *
+   * @param params - Super stream publisher configuration
+   * @param keyExtractor - Function to extract the routing key from a message
+   * @param filter - Optional filter function for server-side message filtering on each partition publisher
+   */
   public async declareSuperStreamPublisher(
     { superStream, publisherRef, routingStrategy }: DeclareSuperStreamPublisherParams,
```
l4mby: Can you change the signature of this method to match `declareSuperStreamPublisher`?
Also please change the jsdoc accordingly

src/client.ts
```diff
@@ -380,7 +380,7 @@ export class Client {
   }
```
l4mby, line 381: Can you also add documentation here?

test/e2e/superstream_consumer.test.ts
```diff
@@ -167,6 +168,45 @@ describe("super stream consumer", () => {
         expect(connections).to.have.length(0)
       }, 5000)
     }).timeout(5000)
+
```
l4mby: This test could be moved into `filtering.test.ts` you could also add some `describes` one for testing classic consumer/publisher and another for super consumer/publisher

test/e2e/superstream_publisher.test.ts
```diff
@@ -339,4 +339,5 @@ describe("super stream publisher", () => {
       }
     }, 5000)
   })
+
```
l4mby: i think you can remove this space

src/client.ts
SumitAthani: Done, updated the method signature and updated the JSDoc as well.

src/client.ts
SumitAthani: Done

test/e2e/superstream_consumer.test.ts
SumitAthani: Done, I've moved the test to `filtering.test.ts` and grouped the tests using separate describe blocks for classic consumer/publisher and super stream consumer/publisher scenarios.

test/e2e/superstream_publisher.test.ts
SumitAthani: Yup :)

l4mby (approved):

